PR #3, opened 2016-07-22
Feature/oauth middleware

italolelis:
## What does this PR do?

This PR adds the possibility to protect routes behind a oauth service.
## What's the authentication flow
1. A authentication request is sent to a authentication route, by default `/oauth/token` with the correct **client_id** and **client_secret** (assuming you are using _client_credentials_)
2. If the request is authorized by the authentication service, the _api-gateway_ will store the access_token session in redis
3. When the client tries to access a protected resource, the _api-geteway_ will try to check in the key-store if the provided token is stored
4. If it is, it will check if the key has expired.
5. In case everything is ok, it will go through
## Impacted Areas in Application

List general components of the application that this PR will affect:
- All protected resources


kieranajp:
:shipit:

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/3/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #9, opened 2016-09-29
Changed to pointer method receiver

italolelis:
## What does this PR do?

 As @endeveit mentioned before it's better using pointer method receivers than literals, so this aims to fix that.

### DIFF
--- a/middleware_oauth.go
+++ b/middleware_oauth.go
@@ -3,17 +3,18 @@ package main
 import (
 	"encoding/json"
 
+	"errors"
+	"net/http"
+
 	log "github.com/Sirupsen/logrus"
 	"github.com/gin-gonic/gin"
-	"net/http"
-	"errors"
 )
 
 type OAuthMiddleware struct {
 	*Middleware
 }
 
-func (m OAuthMiddleware) ProcessRequest(req *http.Request, c *gin.Context) (error, int) {
+func (m *OAuthMiddleware) ProcessRequest(req *http.Request, c *gin.Context) (error, int) {
 	var newSession SessionState
 
 	log.WithFields(log.Fields{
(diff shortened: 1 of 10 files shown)


middleware_oauth.go
```diff
@@ -3,17 +3,18 @@ package main
 import (
 	"encoding/json"
 
+	"errors"
+	"net/http"
```
endeveit, line 7: Good that you grouped your imports in blocks like `stdlib`, `external dependencies`, `project packages` separated with empty line 👍 
But it looks like you forgot to remove empty line on line 5

middleware_oauth2_key_exists.go
```diff
@@ -4,17 +4,18 @@ import (
 	"errors"
 	"strings"
 
+	"net/http"
```
endeveit, line 7: And also here

middleware_oauth.go
italolelis: Actually the empty line is from golint, it puts there automatically

wellfreire:
+1

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/9/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

middleware_oauth.go
endeveit: hm, ok then, but its looks strange for me 🤔

endeveit: Yeah! That is exactly what I'm talking about.

> Imports are organized in groups, with blank lines between them. The standard library packages are in the first group.

And you have an empty line between "encoding/json" and "errors". They are both from standard library

PR #10, opened 2016-09-30
Hotfix/http status code

italolelis:
This fixes problems with wrong HTTP Status codes when we were proxying a request.
Example

`POST users/` Returns `200` instead of `201` even though the auth service was returning `201`


kieranajp:
:shipit:

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/10/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #13, opened 2016-11-07
Feature/local environment

bastiankoetsier:
## What does this PR do?
Setup a local docker-environment for the gateway. It includes the /users endpoint of the auth-service as a fixture. 
Gateway is accessible through port 8888, the auth service itself through 8887. The datastores are exposed through their own ports for debugging purposes.


README.md
```diff
@@ -36,3 +36,12 @@ To start contributing, please check [CONTRIBUTING](CONTRIBUTING.md).
 
 ## Documentation
 * Go lang: https://golang.org/
+
+## Local development
```
wellfreire, line 40: +1 for adding documentation

docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+            context: .
+        volumes:
+            - ./:/go/src/github.com/hellofresh/janus
+        environment:
+            PORT: 3001
+            DEBUG: "true"
+            DATABASE_DSN: "mongodb://mongo:27017"
+            REDIS_DSN: "redis:6379"
+        ports:
+            - "8888:3000"
```
wellfreire, line 14: Shouldn't be `3001` here, since that's the port we have in the env var?

wellfreire (commented):
Apart from that port doubt, LGTM (as I have also tested this locally and it worked).

fixture.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash 
+
```
endeveit, line 2: Could you add a little comment with a description of what this script does?

docker-compose.yml
italolelis: Guys this is an open source repo, we can't have this file here.... it shows the auth-service that is private

italolelis (changes requested):
Guys lets review this, remember that this is an open source repo.....

wellfreire:
@italolelis: You're right. I keep forgetting this is an open source thing. We need to start getting used to that =/

Issue #22, opened 2016-12-18
OAuth Servers ID


italolelis:
Each oauth server has an UUID when created, this id is attached to an API that desires to use that specific oauth server as its auth mechanism. The problem that we have at the moment is that, given the changes on the `Transport` layer, now it's impossible to know from which auth server the access token is coming from, this means that all access_tokens from all servers are on the same `bucket`.  This is not a huge issue when you have just 2 or 3 different oauth servers, but if you have more the access tokens can collide.

PR #23, opened 2016-12-18
Feature/new router

italolelis:
## What does this PR do?
Aims to change the router and middleware provided by gin to a native implementation, this was required because some apis require more customized routing capabilities.

## Mains Changes

**Routing** - The routing system was replaced by [httptreemux](https://github.com/dimfeld/httptreemux) which is an amazing router inspired on [Julien Schmidt's httprouter](https://github.com/julienschmidt/httprouter). This change now brings us more flexibility on the routes definitions because this router is uses the [radix tree](https://en.wikipedia.org/wiki/Radix_tree).

**Middleware** - The middleware chain was also replaced from gin's implementation to [alice](https://github.com/justinas/alice). This package allow us to write `net/http` compatible middleware and chain them in a very nice way.

**Transport** - The transport layer that was defined for our reverse proxy now is smarter and cleaner, I called it `oauth.AwareTransport` since it needs to be aware of the response from the authentication server (so it knows how to store and expire keys automatically).

**Packages** - Most of the structs that were used by janus were in the `janus` package, now they were moved to their own packages, of course this implies in a lot of files changed.

## Known Bugs

**Oauth Servers** - Each oauth server has an UUID when created, this id is attached to an API that desires to use that specific oauth server as its auth mechanism. The problem that we have at the moment is that, given the changes on the `Transport` layer, now it's impossible to know from which auth server the access token is coming from, this means that all access_tokens from all servers are on the same `bucket`.  This is not a huge issue when you have just 2 or 3 different oauth servers, but if you have more the access tokens can collide. I've opened an issue #22 to solve this asap.


bastiankoetsier:
+1

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/23/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #24, opened 2016-12-19
Feature/append listen path

italolelis:
## What does this PR do?
Adds the append listen path logic to it. Now this is the behaviour:

`strip_listen_path`: true - Replaces the listen path in the target path if there is a match
`strip_listen_path`: false - Ignores the listen path and uses only the target path
`append_listen_path`: true - Appends the listen path on the target URL


bastiankoetsier:
+1

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/24/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #38, opened 2016-12-25
Feature/rate limit

italolelis:
## What does this PR do?

- Added [limiter](https://github.com/ulule/limiter) instead of speedbump, since it's more flexible and don't rely only on redis.
- Added a new storage interface
- Added `.editorconfig`
- Refactored the errors


cmd/janus/init.go
```diff
@@ -49,13 +50,21 @@ func init() {
...
+		IdleTimeout: 240 * time.Second,
+		Dial:        func() (redis.Conn, error) { return redis.Dial("tcp", globalConfig.StorageDSN) },
+	}
+
 	dsn := globalConfig.StorageDSN
-	log.Debugf("Trying to connect to redis instance: %s", dsn)
-	redisStorage = redis.NewClient(&redis.Options{
-		Addr: dsn,
-	})
+	log.Debugf("Trying to connect to redis poll: %s", dsn)
```
bastiankoetsier: Typo `poll -> pool`

router/listen_path_matcher.go
```diff
@@ -2,12 +2,16 @@ package router
 
 import "regexp"
 
+var (
+	matchRule = `(\/\*(.+)?)`
```
bastiankoetsier: maybe a `const`?

api/loader.go
```diff
@@ -69,13 +68,19 @@ func (m *Loader) RegisterApi(referenceSpec *Spec) {
...
 
 	if skip {
-		hasher := speedbump.PerSecondHasher{}
-		limit := referenceSpec.RateLimit.Limit
-		limiter := speedbump.NewLimiter(m.redisClient, hasher, limit)
-
 		var handlers []router.Constructor
 		if referenceSpec.RateLimit.Enabled {
-			handlers = append(handlers, limitter.NewRateLimitMiddleware(limiter, hasher, limit).Handler)
+			rate, err := limiter.NewRateFromFormatted(referenceSpec.RateLimit.Limit)
```
bastiankoetsier, line 73: I am not totally sure if this is a breaking change. At least my registered routes for api-v2 (locally) break and my whole app crashes.

bastiankoetsier (changes requested):
It should be clarified that the introduction of the new rate-limit-lib breaks the compatibility to already registered proxies.

api/loader.go
italolelis: Yes It is a breaking change, that's why on the `version` branch we increased the version....

bastiankoetsier:
+1

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/38/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

bastiankoetsier (approved):

PR #47, opened 2017-01-23
Change rate limit redis key prefix to use slug

kieranajp:
Possible fix for rate limit bug assuming redis is having issues with keys containing spaces. Probably good practice even if this isn't the cause of the bug :)


italolelis:
👍 

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/47/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #48, opened 2017-01-23
Verity oauth token origin

vgarvardt:


cmd/janus/main.go
```diff
@@ -68,8 +68,12 @@ func main() {
 	r.Use(middleware.NewLogger(globalConfig.Debug).Handler, middleware.NewRecovery(RecoveryHandler).Handler, middleware.NewMongoDB(accessor).Handler)
 
 	// create the proxy
+	oAuthServersRepo, err := oauth.NewMongoRepository(accessor.Session.DB(""))
+	if err != nil {
+		panic(err)
```
italolelis: Please use `log.Panic`

pkg/oauth/key_exists_middleware.go
```diff
@@ -62,6 +63,17 @@ func (m *KeyExistsMiddleware) Handler(handler http.Handler) http.Handler {
...
 
+		if m.oAuthServerID != thisSessionState.OAuthServerID {
+			log.WithFields(log.Fields{
+				"path":   r.RequestURI,
+				"origin": r.RemoteAddr,
+				"key":    accessToken,
+				"sessionOAuthServerID": thisSessionState.OAuthServerID,
+				"authOAuthServerID":    m.oAuthServerID,
+			}).Info("Attempted access with the key issued by other OAuth provider.")
+			panic(ErrAccessTokenOfOtherOrigin)
```
italolelis, line 75: Please use `log.Panic`

pkg/oauth/mongodb_repository.go
```diff
@@ -101,3 +101,12 @@ func (r *MongoRepository) Remove(id string) error {
 	log.Debugf("Resource %s removed", id)
 	return nil
 }
+
+// FindOAuthServerByTokenURL returns OAuth server records with corresponding token url
+func (r *MongoRepository) FindOAuthServerByTokenURL(url string) (*OAuth, error) {
```
italolelis: Let's call it `FindByTokenURL`

italolelis: Also, don't accept a string, let's accept a `url.URL`

pkg/oauth/transport.go
```diff
@@ -75,6 +78,15 @@ func (t *RoundTripper) RoundTrip(req *http.Request) (resp *http.Response, err er
 		bodyBytes, _ = ioutil.ReadAll(resp.Body)
 
 		if marshalErr := json.Unmarshal(bodyBytes, &newSession); marshalErr == nil {
+			tokenURL := url.URL{Scheme: req.URL.Scheme, Host: req.URL.Host, Path: req.URL.Path}
+			log.WithField("token_url", tokenURL.String()).Debug("Looking for OAuth provider who issued the token")
+			oAuthServer, err := t.oAuthServersRepo.FindOAuthServerByTokenURL(tokenURL.String())
+			if err != nil {
+				log.Error("Failed to find OAuth server by token URL", err)
+				return resp, err
```
italolelis: We can't return here if we do the body on line 95 won't be restored.

pkg/oauth/key_exists_middleware.go
```diff
@@ -19,12 +20,13 @@ var (
 
 // KeyExistsMiddleware checks the integrity of the provided OAuth headers
 type KeyExistsMiddleware struct {
-	manager *Manager
+	manager       *Manager
+	oAuthServerID bson.ObjectId
```
italolelis, line 25: Passing the `api.Spec` pointer is a bit more flexible and probably will allow us to check for other things in the future

italolelis (changes requested):

pkg/oauth/key_exists_middleware.go
vgarvardt: middleware must panic in normal way

italolelis:
👍 

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/48/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

italolelis (approved):

PR #66, opened 2017-02-06
Replace IDs with placeholder

vgarvardt:
Replace IDs with placeholder to avoid huge load of useless path metrics


pkg/stats/bucket.go
```diff
@@ -4,14 +4,46 @@ import (
...
 const (
-	TotalRequestBucket   = "total.requests"
-	TotalRoundTripBucket = "total.round"
+	totalRequestBucket   = "total.requests"
+	totalRoundTripBucket = "total.round"
+	pathIDPlaceholder    = "-id-"
 )
 
+func testAlwaysTrue(string) bool {
+	return true
```
kieranajp, line 18: :suspect:

vgarvardt: This is test function stub for paths that always has ID at second level (most of paths actually). See `hasIDAtSecondLevel` bellow.

kieranajp:
:shipit:

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/66/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #71, opened 2017-02-07
Added store unit sets

vgarvardt:


pkg/store/redis.go
```diff
@@ -22,6 +22,9 @@ type RedisStore struct {
 
 	// The maximum number of retry under race conditions.
 	MaxRetry int
+
+	// Redis connection stub for tests
+	mockGetConnection func() redis.Conn
```
italolelis: This really doesn't feel right... having a mocking method on a struct.
Probably better to have an `TestingStore` where you can mock the `GetConnection` method

italolelis (changes requested):

pkg/store/redis.go
vgarvardt: Removed mock method and extracted all logic to private methods that are tested now.

vgarvardt:
Updated PR to reflect the latest changes from https://github.com/hellofresh/janus/pull/85

italolelis (approved):

italolelis:
👍 

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/71/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #73, opened 2017-02-09
Fixed 301 behavior for paths with slash at the end

vgarvardt:


vgarvardt:
affects only tests, not critical, but fix may break tons of live code

PR #75, opened 2017-02-09
Fixes problems with long expiration time

italolelis:
## What does this PR do?

Fixes problems dealing with long numbers being converted to seconds which makes them even bigger.
When working with, let's say a month of expiration time which is `2629743` seconds, go was converting this to the scientific notation `2.592e+6`, which is 2 seconds for redis.

This PR fixes this problem by only accepting integers as a value for expiration

### DIFF
--- a/pkg/oauth/manager.go
+++ b/pkg/oauth/manager.go
@@ -2,7 +2,6 @@ package oauth
 
 import (
 	"encoding/json"
-	"time"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


kieranajp:
+1

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/75/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #77, opened 2017-02-16
Track muted stats in logs

vgarvardt:
After merging I suggest to disable stats collecting in teams staging and, maybe, on global staging as it takes to much space in graphite - today I removed `7.5GB` staging counters and `4.2GB` staging timers data from graphite as it run out of free space again.

### DIFF
--- a/cmd/janus/init.go
+++ b/cmd/janus/init.go
@@ -26,7 +26,7 @@ var (
 func init() {
 	globalConfig, err = config.LoadEnv()
 	if nil != err {
-		log.Panic(err.Error())
+		log.WithError(err).Panic("Failed to load config from environment")
 	}
 }
 
@@ -34,7 +34,7 @@ func init() {
 func init() {
 	level, err := log.ParseLevel(strings.ToLower(globalConfig.LogLevel))
 	if err != nil {
-		log.Error("Error getting level", err)
+		log.WithError(err).Error("Error getting level")
 	}
 
 	log.SetLevel(level)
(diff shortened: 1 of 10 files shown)


cmd/janus/main.go
```diff
@@ -65,7 +65,7 @@ func main() {
 	defer accessor.Close()
 	defer statsdClient.Close()
 
-	statsClient := stats.NewStatsClient(statsdClient)
+	statsClient := stats.NewStatsClient(statsdClient, globalConfig.Statsd.DSN == "")
```
italolelis: I understand the motivation of doing this and I agree that we should have it, just the implementation that is not really nice. If we create an interface for the `StatsClient` we could have the real `StatsClient` and if `globalConfig.Statsd.DSN == ""` we would use the `MutedClient` that only logs (strategy pattern), this way we wouldn't need to change the contract to have mute option that makes it not really clear.

italolelis (changes requested):

cmd/janus/main.go
vgarvardt: I want Live and Muted clients to produce the same metrics, so I generalised incrementer and time tracker.

vgarvardt: Incapsulated statsd into stats completely

cmd/janus/init.go
```diff
@@ -64,19 +65,21 @@ func init() {
...
 func init() {
 	var options []statsd.Option
 	statsdConfig := globalConfig.Statsd
+	muted := false
 
 	log.Debugf("Trying to connect to statsd instance: %s", statsdConfig.DSN)
 	if len(statsdConfig.DSN) == 0 {
 		log.Debug("Statsd DSN not provided, client will be muted")
 		options = append(options, statsd.Mute(true))
+		muted = true
```
italolelis, line 82: If you do `statsd.Mute` you already have the value... I think the `muted` extra var is unecessary

italolelis (changes requested):

vgarvardt:
Extracted this code to separate `stats` package. Will be released soon and will replace stats package in the code base.

PR #85, opened 2017-02-26
Feature/revoke

italolelis:
## What does this PR do?

Adds the capability to revoke tokens from the chosen storage (redis in our case). With this, you make sure that your storage is synchronized with the authentication provider. 
The ideal scenario though is to use low expiration tokens, this way you wouldn't need to revoke a token manually.


vgarvardt:
👍 

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/85/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #90, opened 2017-03-01
PLAT-263 Feature/jwt check

italolelis:
## What does this PR do?

- Adds the ability to choose which storage to use
- Adds token strategy, this way you can choose which one makes more sense
- Better error messages
- Better logging

hellofresh/automation#962 should be merged before this one.


pkg/oauth/manager_factory.go
```diff
@@ -0,0 +1,70 @@
...
+	case "storage":
+		return Storage, nil
+	case "jwt":
+		return JWT, nil
+	case "auth":
+		return Auth, nil
+	}
+
+	var m ManagerType
+	return m, ErrUnknownStrategy
```
vgarvardt: `return nil, ...`

vgarvardt:
👍 

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/90/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

Issue #91, opened 2017-03-01
Dockerfile based build : Empty reply from server


drauschenbach:
#I can't seem to get the Docker-based build to respond to any http requests. I'm using a build of the latest [1.7.4-rc.6](https://github.com/hellofresh/janus/releases/tag/1.7.4-rc.6) using my modified [Dockerfile](https://github.com/Megalithic-LLC/janus-dockerfile), that otherwise matches the standard one but allows me to specify a tag to build.

**Reproduction Steps:**

1. Build

```bash
$ git clone https://github.com/Megalithic-LLC/janus-dockerfile.git
$ cd janus-dockerfile
$ docker build --build-arg=JANUS_TAG=1.7.4-rc.7 -t janus .
```

2. Configure runtime env vars:

```bash
$ vi janus.env

DATABASE_DSN=mongodb://1.2.3.4/janus
REDIS_DSN=redis://1.2.3.4:6379
SECRET=password
ADMIN_USERNAME=admin
ADMIN_PASSWORD=admin
DEBUG=true
LOG_LEVEL=debug
```

3. Launch Janus

```bash
$ docker run --env-file janus.env -i -p 8080:8080 -t janus
{"@timestamp":"2017-03-01T17:44:59.246556501Z","@version":"1","level":"debug","message":"Connected to mongodb","type":"Janus"}
...
01T17:44:59.250617897Z","@version":"1","level":"info","message":"Listening on :0","type":"Janus"}
```

4. Try and talk to it

```
$ curl -iv -X POST -H "Content-Type: application/json" -H "Accept: application/json, */*" -d '{"username":"admin","password":"admin"}' http://localhost:8080/login

*   Trying ::1...
* Connected to localhost (::1) port 8080 (#0)
> POST /login HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.43.0
> Content-Type: application/json
> Accept: application/json, */*
> Content-Length: 39
> 
* upload completely sent off: 39 out of 39 bytes
* Empty reply from server
* Connection #0 to host localhost left intact

curl: (52) Empty reply from server
```

**Janus version:** [1.7.4-rc.6](https://github.com/hellofresh/janus/releases/tag/1.7.4-rc.6)
**OS and version:** standard Dockerfile (golang:1.8-alpine)

italolelis:
Hi @drauschenbach,

[...]

PR #92, opened 2017-03-01
Fixes the default value for the port

italolelis:
## What does this PR do?

This fixes the problems reported on #91 by setting a default port value to the config which matches the one set on the image.


kieranajp:
:shipit:

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/92/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #109, opened 2017-03-20
Implement OpenTracing

italolelis:
## What does this PR do?

Implements [opentracing](http://opentracing.io/) for janus. You can choose between 2 tracers:

* Google Cloud Platform - [Stackdrive trace](https://cloud.google.com/trace/)
* [Appdash](https://github.com/sourcegraph/appdash) - Local or Remote

If you choose appdash as your tracer you can choose to run it using a shared instance (remote), in this case, you need to have an appdash server somewhere and point Janus to it. 
You can also use the default implementation that Janus provide with a built in appdash server.

I also added some basic tracing for the api and auth endpoints

### DIFF
--- a/pkg/api/handler.go
+++ b/pkg/api/handler.go
@@ -24,9 +24,7 @@ func NewController(repo Repository) *Controller {
 func (c *Controller) Get() http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		span := opentracing.FromContext(r.Context(), "datastore.FindAll")
-
 		data, err := c.repo.FindAll()
-
 		span.Finish()
 
 		if err != nil {
--- a/pkg/api/handler.go
+++ b/pkg/api/handler.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 
 	"github.com/hellofresh/janus/pkg/errors"
+	"github.com/hellofresh/janus/pkg/opentracing"
 	"github.com/hellofresh/janus/pkg/request"
 	"github.com/hellofresh/janus/pkg/response"
 	"github.com/hellofresh/janus/pkg/router"
@@ -22,7 +23,12 @@ func NewController(repo Repository) *Controller {
 
 func (c *Controller) Get() http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
+		span := opentracing.FromContext(r.Context(), "datastore.FindAll")
+
 		data, err := c.repo.FindAll()
+
+		span.Finish()
+
 		if err != nil {
 			panic(err.Error())
 		}
@@ -34,8 +40,10 @@ func (c *Controller) Get() http.HandlerFunc {
 func (c *Controller) GetBy() http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		name := router.FromContext(r.Context()).ByName("name")
-
+		span := opentracing.FromContext(r.Context(), "datastore.FindByName")
 		data, err := c.repo.FindByName(name)
+		span.Finish()
+
 		if data == nil {
 			panic(ErrAPIDefinitionNotFound)
 		}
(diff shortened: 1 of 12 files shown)


pkg/api/handler.go
```diff
@@ -22,7 +23,12 @@ func NewController(repo Repository) *Controller {
 
 func (c *Controller) Get() http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
+		span := opentracing.FromContext(r.Context(), "datastore.FindAll")
+
```
vgarvardt: extra new line is not required

pkg/api/handler.go
```diff
@@ -22,7 +23,12 @@ func NewController(repo Repository) *Controller {
...
 		data, err := c.repo.FindAll()
+
```
vgarvardt: extra new line is not required

pkg/config/specification.go
```diff
@@ -70,6 +60,36 @@ type Credentials struct {
...
+
+// Tracing represents the distributed tracing configuration
+type Tracing struct {
+	GoogleCloudTracing GoogleCloudTracing
+	AppdashTracing     AppdashTracing
+}
+
+// IsGoogleCloudEnabled checks if google cloud is enabled
+func (t Tracing) IsGoogleCloudEnabled() bool {
+	return len(t.GoogleCloudTracing.Email) > 0 && len(t.GoogleCloudTracing.PrivateKey) > 0 && len(t.GoogleCloudTracing.PrivateKeyID) > 0
```
vgarvardt: don't we need to check `ProjectID` either?

pkg/web/provider.go
```diff
@@ -64,7 +65,8 @@ func (p *Provider) listenAndServe(handler http.Handler) error {
 		return http.ListenAndServeTLS(address, p.CertFile, p.KeyFile, handler)
 	}
 
-	log.Infof("certPathTLS or keyPathTLS not found, defaulting to HTTP")
+	log.Info("Certificate and certificate key were not found, defaulting to HTTP")
+	log.Info("Janus Admin API started")
```
vgarvardt: This log message exists only for HTTP, but not for HTTPS. Also, log message should have static string and context for message above (can not leave a comment there), something like `log.WithField("address", address).Info("Listening Admin API @")`.

Same applies to main web server.

pkg/config/specification.go
```diff
@@ -8,32 +8,22 @@ import (
...
+	InsecureSkipVerify   bool          `envconfig:"INSECURE_SKIP_VERIFY" description:"Disable SSL certificate verification"`
+	StorageDSN           string        `envconfig:"STORAGE_DSN" default:"memory://localhost" description:"The Storage DSN, this could be 'memory' or 'redis'"`
+	CertPathTLS          string        `envconfig:"CERT_PATH" description:"Path of certificate when using TLS"`
+	KeyPathTLS           string        `envconfig:"KEY_PATH" description:"Path of key when using TLS"`
+	BackendFlushInterval time.Duration `envconfig:"BACKEND_FLUSH_INTERVAL" default:"20ms" description:"Flush interval for upgraded Proxy connections"`
+	CloseIdleConnsPeriod time.Duration `envconfig:"CLOSE_IDLE_CONNS_PERIOD" description:"Defines the time period of how often the idle connections maintained by the proxy are closed."`
+	Database             Database
+	Statsd               Statsd
+	Credentials          Credentials
+	Tracing              Tracing
```
vgarvardt, line 26: We need some docs on this. I think adding this to `Misc` section as I did for Monitoring should be fine.

vgarvardt (changes requested):

pkg/config/specification.go
italolelis: I'll do documentation on a separated PR, right now I just want to put this to staging and auth to test it out

vgarvardt (approved):

vgarvardt:
👍 

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/109/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #111, opened 2017-03-21
Remove glide cache from image to dramatically reduce docker image size

vgarvardt:
I found that docker image dramatically increased after tracing feature merge. Investigation shown that glide cache has significant size and after application build we do not need it at all.

```
$ docker run janus:dev du -hs /root/.glide
246.4M	/root/.glide
```


italolelis:
👍 

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/111/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #117, opened 2017-03-23
Merged PR pipeline into main pipeline

vgarvardt:


italolelis:
Why did u do this change at all? I believe this is not what we want....
It was like this in the beginning but having different pipelines is always better to visualize than having everything in a single `pipeline` file. We should actually have the same kind of splitting in `auth-service`.
For now, I'm splitting back into 2 separate pipelines in concourse since there is no reason to have this.

italolelis:
👎

vgarvardt:
Got your point.

PR #118, opened 2017-03-29
New improvements

italolelis:
## What does this PR do?

* Adds tests for the critical path of the app. Basically, it tests the proxy rules.
* Adds plugins capabilities. With this, we don't need to couple the `RoundTrip`  with everyone that wants to change the response from the proxy. Now we just need to add a plugin that will be called in a chain when the response is back.
* Adds multiple ways to provide configuration to Janus. Now you can choose if you wanna use `environment variables`, `YAML`, `JSON` or `TOML`.
* Updated docker compose to use the `TOML` config file
* Using commands to start Janus. This way we can improve the organization on how we want the binary to work. Also, this will allow us to probably move towards an ideal solution for hot reload of configs.
* Added coverage on TravisCI

This PR doesn't break compatibility.


.travis.yml
```diff
@@ -11,7 +11,11 @@ before_install:
...
 install:
   - mkdir -p $GOPATH/bin
+  - go get github.com/mattn/goveralls
+  - go get github.com/go-playground/overalls
   - make deps
 
 script:
 - make test
+- $GOPATH/bin/overalls -project=github.com/hellofresh/janus -covermode=count
+- $GOPATH/bin/goveralls -coverprofile=overalls.coverprofile -service=travis-ci -repotoken $COVERALLS_TOKEN
```
vgarvardt, line 21: 👍

pkg/proxy/reverseproxy.go
```diff
@@ -69,18 +71,17 @@ func WithParams(o Params) *Proxy {
...
-		MaxIdleConns:          100,
-		IdleConnTimeout:       90 * time.Second,
-		TLSHandshakeTimeout:   10 * time.Second,
-		ExpectContinueTimeout: 1 * time.Second,
-		MaxIdleConnsPerHost:   o.IdleConnectionsPerHost,
-	}
+	tr := http.DefaultTransport.(*http.Transport)
+	tr.Proxy = http.ProxyFromEnvironment
+	tr.DialContext = (&net.Dialer{
+		Timeout:   30 * time.Second,
```
vgarvardt, line 77: I think it would be nice to have low-level proxy settings configurable with default values set to these. Not a must, but nice to have.

pkg/proxy/reverseproxy_test.go
```diff
@@ -0,0 +1,116 @@
...
+		assert.Equal(t, tc.expectedContentType, res.Header.Get("Content-Type"))
+		assert.Equal(t, tc.expectedCode, res.StatusCode, tc.description)
+	}
+}
+
+func createProxyDefinitions() []*proxy.Definition {
+	return []*proxy.Definition{
+		&proxy.Definition{
+			ListenPath:  "/example/*",
+			UpstreamURL: "http://www.mocky.io/v2/58c6c60710000040151b7cad",
```
vgarvardt, line 70: It's definitely bad idea to make real http requests in unit tests. I understand that this is really useful thing, but not part of unit tests. Let's think about introducing integration tests here and run them separately.

pkg/proxy/shackles.go
```diff
@@ -0,0 +1,109 @@
...
+type OutLink func(req *http.Request, res *http.Response) (*http.Response, error)
+
+// InChain typed array for inbound plugin sequence
+type InChain []InLink
+
+// OutChain typed array for outbound plugin sequence
+type OutChain []OutLink
+
+// Shackles construct holding plugin sequences
+type Shackles struct {
```
vgarvardt, line 27: nice name =)

pkg/proxy/shackles.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// RoundTrip provides the Transport.RoundTrip function to handle requests the proxy receives
+func (s *Shackles) RoundTrip(req *http.Request) (resp *http.Response, err error) {
+	timing := s.statsClient.BuildTimeTracker().Start()
+
+	// apply inbound request plugins (if any)
+	req, err = s.applyInboundLinks(req)
+	if err != nil {
+		return nil, err
```
vgarvardt: We must monitor errors like this - currently you're monitoring (using stats) only http-request fail, but any of the errors here make proxy request fail - `s.applyInboundLinks()`,`httputil.DumpResponse()`, `s.applyOutboundLinks()`

vgarvardt (changes requested):

pkg/proxy/reverseproxy.go
italolelis: Yes, we can do that, but not part of this PR.

pkg/proxy/reverseproxy_test.go
italolelis: Yes, I agree. Just remember one thing, as we act as a proxy we need something to proxy to, and that is probably not under our control. But I agree it should be done as an integration test.

vgarvardt (approved):

vgarvardt:
👍 

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/118/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #123, opened 2017-04-10
Inject app version in compile time

vgarvardt:
**BEFORE MERGE**: apply pipeline changes

### DIFF
--- a/cmd/janus/server.go
+++ b/cmd/janus/server.go
@@ -30,16 +30,6 @@ var (
 
 // RunServer is the run command to start Janus
 func RunServer(cmd *cobra.Command, args []string) {
-	var printVersion bool
-	printVersion, err = cmd.Flags().GetBool(flagVersion)
-	if nil != err {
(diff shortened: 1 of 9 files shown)


cmd/janus/server.go
```diff
@@ -30,7 +30,17 @@ var (
 
 // RunServer is the run command to start Janus
 func RunServer(cmd *cobra.Command, args []string) {
-	log.Info("Janus starting...")
+	var printVersion bool
+	printVersion, err = cmd.Flags().GetBool(flagVersion)
```
italolelis: This should be in a separated command. Something like `janus version`

vgarvardt: fixed, will do the same for auth-service if this solution is fine

italolelis:
👍 

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/123/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #138, opened 2017-05-08
Make api respond not found

rafaeljesus:
## What does this PR do?
The janus API was responding 500 when no proxy found. This PR fix it.

### DIFF
--- a/pkg/api/mongodb_repository.go
+++ b/pkg/api/mongodb_repository.go
@@ -51,6 +51,10 @@ func (r *MongoRepository) FindByName(name string) (*Definition, error) {
 	defer session.Close()
 
 	err := coll.Find(bson.M{"name": name}).One(&result)
+	if err != nil {
+		return nil, ErrAPIDefinitionNotFound
+	}
+
 	return result, err
 }
 
--- a/pkg/api/mongodb_repository.go
+++ b/pkg/api/mongodb_repository.go
@@ -52,7 +52,10 @@ func (r *MongoRepository) FindByName(name string) (*Definition, error) {
 
 	err := coll.Find(bson.M{"name": name}).One(&result)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/api/mongodb_repository.go
```diff
@@ -51,6 +51,10 @@ func (r *MongoRepository) FindByName(name string) (*Definition, error) {
 	defer session.Close()
 
 	err := coll.Find(bson.M{"name": name}).One(&result)
+	if err != nil {
+		return nil, ErrAPIDefinitionNotFound
```
vgarvardt: this can be mongo connection error - is there a way to check if this is real error or result error?

vgarvardt (changes requested):

pkg/api/mongodb_repository.go
rafaeljesus: I will try

pkg/api/mongodb_repository.go
```diff
@@ -51,6 +51,13 @@ func (r *MongoRepository) FindByName(name string) (*Definition, error) {
 	defer session.Close()
 
 	err := coll.Find(bson.M{"name": name}).One(&result)
+	if err != nil {
+		if err == mgo.ErrNotFound {
```
rafaeljesus, line 55: @vgarvardt wdyt?

rafaeljesus: https://github.com/go-mgo/mgo/blob/v2/session.go#L2330

vgarvardt (approved):

vgarvardt:
👍 

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/138/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #143, opened 2017-05-20
Allow to register a proxy with a ListenPath-less configuration.

sroze:
So we can now register proxies without a listen path:

```go
register.Add(proxy.NewRoute(
    &proxy.Definition{
        PreserveHost: true,
        UpstreamURL: "https://the-backend",
        Methods: []string{"ALL"},
    },
    middleware.NewHostMatcher([]string{"my-hostname.com"}).Handler,
))
```

I'm not sure how to write a test for this one.


italolelis:
Hi @sroze,

Thank you for your PR. Could explain a bit more how this would be useful to have? 
For what I understand all endpoints should have a listen_path, otherwise you don't know how to proxy. Also, this can lead to collisions of routes if you add 2 proxies without listen_path.

sroze:
Hi @italolelis! This capability can be used if you can using the host matcher and are sure to have only one wildcard rule for a domain name. For instance, I know that for the `api-specific.my-hostname.com` domain name, I'll forward everything to the given upstream.

italolelis:
@sroze Got it. But the current implementation won't do that. The chi router doesn't take into consideration the host name when allocating routes in the tree, meaning that we will have a conflict when you add two routes with two different hosts.
If we'd like to add this feature, we might need to change how chi works (which I think it's a lot of work)

italolelis:
Closing this PR for inactivity. Please feel free to reopen it if necessary.

PR #144, opened 2017-05-21
Allow to configure the user key in JWT token

sroze:
If the user identifier key is not `id` in the JWT token, it is important to be able to configure which one should be used.

**Backward compatibility break**: the default value being `sub` has per the JWT specification, it will probably break setups using the `id` key only.


italolelis:
Hi @sroze. Can u please gofmt this file `pkg/jwt/middleware.go`? CI is failing because of that.

sroze:
@italolelis done :)

italolelis:
@sroze Please check CI, there are some lint rules that are breaking.

italolelis:
Closing this PR for inactivity. Please feel free to reopen it if necessary.

PR #188, opened 2017-08-30
Added GitHub login

italolelis:
## What does this PR do?

Adds possibility to login to the admin API with github account

### DIFF
--- a/docs/quick_start/add_endpoint.md
+++ b/docs/quick_start/add_endpoint.md
@@ -32,7 +32,7 @@ You can also configure which organizations/teams will be allowed to login on the
 {% codetabs name="HTTPie", type="bash" -%}
 http -v --json POST localhost:8081/login?provider=github "Authorization:Bearer githubToken"
 {%- language name="CURL", type="bash" -%}
-http -X "POST" localhost:8081/login?provider=github -H 'Authorization:Bearer githubToken'
+curl -X "POST" localhost:8081/login?provider=github -H 'Authorization:Bearer githubToken'
 {%- endcodetabs %}
 
 ### Basic
@@ -52,7 +52,7 @@ With the basic auth provider you can easily login to janus admin API with `Basic
 {% codetabs name="HTTPie", type="bash" -%}
 http -v --json POST localhost:8081/login username=admin password=admin
 {%- language name="CURL", type="bash" -%}
-http -X "POST" localhost:8081/login -d '{"username": "admin", "password": "admin"}'
+curl -X "POST" localhost:8081/login -d '{"username": "admin", "password": "admin"}'
 {%- endcodetabs %}
 
 The username and password defaults to *admin*/*admin*.
--- a/docs/quick_start/add_endpoint.md
+++ b/docs/quick_start/add_endpoint.md
@@ -6,10 +6,48 @@ The main feature of the API Gateway is to proxy the requests to different servic
 
 ## Authenticating
 
-To start using the Janus adminstration API you need to get a [JSON Web Token](https://jwt.io) and provide it in every single request
-using the `Authorization` header.
+To start using the Janus adminstration API you need to get a [JSON Web Token](https://jwt.io) and provide it in every single request using the `Authorization` header.
 
-To get a token you must execute:
+You can choose between login with `github` or `basic` providers.
+
+### Github
+
+To login with github you need to send a valid github access token in the Authorization header. This token will be exchanged by a JWT token that you can use to make requests to the gateway gateway API.
+
+You can choose to either go through the [oAuth2](https://developer.github.com/v3/guides/basics-of-authentication/) flows to authorize an user on github or generate a [Personal Access Toke](https://github.com/settings/tokens)
+
+You can also configure which organizations/teams will be allowed to login on the Admin API. For this you just need to configure:
+
+```toml
+  [web.credentials]
+    secret = "secret"
+
+    [web.credentials.github]
+    organizations = ["hellofresh"]
+    teams = [
+      {organizationName = "hellofresh", TeamName = "Devs"}
+    ]
+```
+
+{% codetabs name="HTTPie", type="bash" -%}
+http -v --json POST localhost:8081/login?provider=github "Authorization:Bearer githubToken"
+{%- language name="CURL", type="bash" -%}
+http -X "POST" localhost:8081/login?provider=github -H 'Authorization:Bearer githubToken'
+{%- endcodetabs %}
+
+### Basic
+
+With the basic auth provider you can easily login to janus admin API with `Basic` Authentication, sending a JSON with the credentials or Form parameters.
+
+```toml
+  [web.credentials]
+    secret = "secret"
+
+    [web.credentials.basic]
+    users = [
+      {username = "admin", password = "admin"}
+    ]
+```
 
 {% codetabs name="HTTPie", type="bash" -%}
 http -v --json POST localhost:8081/login username=admin password=admin
@@ -17,7 +55,7 @@ http -v --json POST localhost:8081/login username=admin password=admin
 http -X "POST" localhost:8081/login -d '{"username": "admin", "password": "admin"}'
 {%- endcodetabs %}
 
-The username and password are defined by the configuration called `web.credentials.username` and `web.credentials.password`. It defaults to *admin*/*admin*.
+The username and password defaults to *admin*/*admin*.
 
 <p align="center">
   <a href="http://g.recordit.co/dDjkyDKobL.gif">
--- a/docs/quick_start/add_endpoint.md
+++ b/docs/quick_start/add_endpoint.md
@@ -12,9 +12,9 @@ You can choose between login with `github` or `basic` providers.
 
 ### Github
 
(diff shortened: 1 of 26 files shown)


pkg/jwt/basic/password_verifier.go
```diff
@@ -0,0 +1,81 @@
...
+// NewPasswordVerifier creates a new instance of PasswordVerifier
+func NewPasswordVerifier(users []*User) *PasswordVerifier {
+	return &PasswordVerifier{users}
+}
+
+// Verify makes a check and return a boolean if the check was successful or not
+func (v *PasswordVerifier) Verify(r *http.Request) (bool, error) {
+	currentUser, err := v.getUserFromRequest(r)
+	if err != nil {
+		log.Debug("Could not get user from request")
```
rafaeljesus, line 40: shouldn't we be returning `false, nil` here?

docs/quick_start/add_endpoint.md
```diff
@@ -6,18 +6,56 @@ The main feature of the API Gateway is to proxy the requests to different servic
...
-To start using the Janus adminstration API you need to get a [JSON Web Token](https://jwt.io) and provide it in every single request
-using the `Authorization` header.
+To start using the Janus adminstration API you need to get a [JSON Web Token](https://jwt.io) and provide it in every single request using the `Authorization` header.
 
-To get a token you must execute:
+You can choose between login with `github` or `basic` providers.
+
+### Github
+
+To login with github you need to send a valid github access token in the Authorization header. This token will be exchanged by a JWT token that you can use to make requests to the gateway gateway API.
```
rafaeljesus: `gateway gateway API`

docs/quick_start/add_endpoint.md
```diff
@@ -6,18 +6,56 @@ The main feature of the API Gateway is to proxy the requests to different servic
...
+
+You can choose to either go through the [oAuth2](https://developer.github.com/v3/guides/basics-of-authentication/) flows to authorize an user on github or generate a [Personal Access Toke](https://github.com/settings/tokens)
```
rafaeljesus: `Toke`

pkg/jwt/basic/password_verifier.go
```diff
@@ -0,0 +1,82 @@
...
+	username, password, ok := r.BasicAuth()
+	if ok {
+		user = &User{
+			Username: username,
+			Password: password,
+		}
+	}
+
+	// checks if the content is json otherwise just get from the form params
+	if r.Header.Get("Content-Type") == "application/json" {
```
rafaeljesus: Maybe extract a const `r.Header.Get("Content-Type") == contentTypeJSON`

pkg/jwt/basic/provider.go
```diff
@@ -0,0 +1,35 @@
...
+		Verifier: provider.NewVerifierBasket(
+			NewPasswordVerifier(userConfigToTeam(config.Basic.Users)),
+		),
+	}
+}
+
+func userConfigToTeam(configUser []config.BasicUsersConfig) []*User {
+	users := []*User{}
+	for _, u := range configUser {
+		users = append(users, &User{
```
rafaeljesus: maybe it's worth to extract a constructor `NewUser(username, password)`

pkg/jwt/github/client.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return *currentUser.Login, nil
+}
+
+// Teams retrieves the teams that the authenticated user belongs
+func (c *client) Teams(httpClient *http.Client) (OrganizationTeams, error) {
+	client := github.NewClient(httpClient)
```
rafaeljesus, line 43: This could have being created on the constructor `NewClient(httpClient *http.Client)`

pkg/jwt/basic/provider.go
```diff
@@ -0,0 +1,35 @@
...
+	provider.Register("basic", Provider{})
+}
+
+// Provider abstracts the authentication for github
+type Provider struct {
+	provider.Verifier
+}
+
+// Build acts like the constructor for a provider
+func (gp Provider) Build(config config.Credentials) provider.Provider {
```
rafaeljesus, line 18: I'd rather `New` over `Build` but if this convention is being used that's fine

pkg/jwt/github/organization_verifier.go
```diff
@@ -0,0 +1,50 @@
...
+// NewOrganizationVerifier creates a new instance of OrganizationVerifier
+func NewOrganizationVerifier(organizations []string, gitHubClient Client) *OrganizationVerifier {
+	return &OrganizationVerifier{
+		organizations: organizations,
+		gitHubClient:  gitHubClient,
+	}
+}
+
+// Verify makes a check and return a boolean if the check was successful or not
+func (verifier OrganizationVerifier) Verify(r *http.Request) (bool, error) {
```
rafaeljesus: `verifier *OrganizationVerifier`

pkg/jwt/github/team_verifier.go
```diff
@@ -0,0 +1,58 @@
...
+// NewTeamVerifier creates a new instance of TeamVerifier
+func NewTeamVerifier(teams []Team, gitHubClient Client) *TeamVerifier {
+	return &TeamVerifier{
+		teams:        teams,
+		gitHubClient: gitHubClient,
+	}
+}
+
+// Verify makes a check and return a boolean if the check was successful or not
+func (verifier TeamVerifier) Verify(r *http.Request) (bool, error) {
```
rafaeljesus: `verifier *TeamVerifier`

pkg/jwt/handler.go
```diff
@@ -24,36 +24,30 @@ type Login struct {
...
 	return func(w http.ResponseWriter, r *http.Request) {
-		var loginValues Login
+		factory := provider.Factory{}
+		p := factory.Build(r.URL.Query().Get("provider"), config)
 
-		if json.NewDecoder(r.Body).Decode(&loginValues) != nil {
-			j.Guard.Unauthorized(w, r, errors.New("missing username or password"))
+		verified, err := p.Verify(r)
+		if err != nil {
+			render.JSON(w, http.StatusInternalServerError, err.Error())
```
rafaeljesus: Are you sure this need to be `500`?

pkg/jwt/provider/verifier_test.go
```diff
@@ -0,0 +1,97 @@
...
+func (v *mockVerifier) Verify(r *http.Request) (bool, error) {
+	return v.result1, v.result2
+}
+
+func (v *mockVerifier) VerifyReturns(result1 bool, result2 error) {
+	v.result1 = result1
+	v.result2 = result2
+}
+
+func TestAllVerifiersFailed(t *testing.T) {
```
rafaeljesus: We could start using table tests, we have a example on push-notification-service

pkg/jwt/basic/provider.go
italolelis: It's not worth it, the user is only used by this package and only by this function. A constructor should only be used if:

- It's gonna be exposed
- It requires some contextual validation to avoid inconsistent objects

pkg/jwt/github/client.go
italolelis: No, it couldn't. The constructor is called on start up time, the Teams is called on each authentication requests with different `http` contexts, which means you have to create a new GitHub client for each request that will contain different contexts

pkg/jwt/basic/provider.go
italolelis: The provider is acting as an [Abstract Factory](https://sourcemaking.com/design_patterns/abstract_factory) where it will create an object based on the given configurations. Do you think it should be `New` anyway?

pkg/jwt/basic/provider.go
rafaeljesus: Agreed, if it's not being consumed from outside that's right, as I saw the user struct and attributes publicly declared I thought it would be. Maybe its makes sense to declare it private then

pkg/jwt/github/client.go
rafaeljesus: Got it 👍

pkg/jwt/basic/provider.go
rafaeljesus: Hm now I see you are right, better keep any `new` inside of the factory, thank you!

pkg/jwt/provider/provider.go
```diff
@@ -6,29 +6,25 @@ import (
 	"github.com/hellofresh/janus/pkg/config"
 )
 
-var providers map[string]Provider
-var mutex sync.Mutex
+var providers *sync.Map
+
+func init() {
+	providers = new(sync.Map)
```
rafaeljesus: 🤙

rafaeljesus:
👍 

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/188/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway) [Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/188/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

PR #215, opened 2017-10-11
add prometheus metrics support for proxy

alileza:
## What does this PR do?
Add [prometheus](github.com/prometheus/prometheus) metrics support


pkg/proxy/transport.go
```diff
@@ -46,6 +47,8 @@ type Params struct {
 	FlushInterval time.Duration
 
 	Outbound OutChain
+
+	Prometheus config.Prometheus
```
italolelis: Why do we need to inject this here?

pkg/proxy/register.go
```diff
@@ -54,7 +55,17 @@ func (p *Register) Add(route *Route) error {
 		p.doRegister(matcher.Extract(definition.ListenPath), handler.ServeHTTP, definition.Methods, route.Inbound)
 	}
 
-	p.doRegister(definition.ListenPath, handler.ServeHTTP, definition.Methods, route.Inbound)
+	handle := handler.ServeHTTP
+	if p.params.Prometheus.Enabled {
```
italolelis: I would put this in a plugin that can based on the configurations, enable or disable Prometheus.

pkg/web/provider.go
```diff
@@ -62,6 +64,10 @@ func onStartup(event interface{}) error {
 	r.GET("/status", checker.NewOverviewHandler(repo))
 	r.GET("/status/{name}", checker.NewStatusHandler(repo))
 
+	if e.Config.Prometheus.Enabled {
```
italolelis: I would move this to a plugin as listen to the events to register a new route.

italolelis (changes requested):
Hi @alileza,

Thank you for your great contribution. I have some suggestions about the implementation though.

To be less intrusive on the code, we could build this functionality as a plugin. We could listen to the events that Janus triggers and then register the routes, middlewares and fetch configurations easily.

You can perhaps take a look how other plugins are made and do some refactoring?

Thanks,

alileza:
Hi @italolelis,

Thank's for the quick review, here's I update the PR to act as a plugin instead of messing with the core code. 

The reason that I didn't use an event to enable prometheus default metric path is, because it's because it require addition on `provider` server and might need custom `metric_path`, which should be specify by user. Please do add some advise if there's a better way to do it.

Thank :)

pkg/web/provider.go
```diff
@@ -61,6 +62,9 @@ func onStartup(event interface{}) error {
 	// health checks
 	r.GET("/status", checker.NewOverviewHandler(repo))
 	r.GET("/status/{name}", checker.NewStatusHandler(repo))
+	if e.Config.Prometheus.Enabled {
```
italolelis: This can be moved to the plugin. The `plugin.OnStartup` event contains the Global configurations for Janus and this event runs when Janus starts. On the `plugin.OnAdminAPIStartup` you have access to the admin router to add this route.

italolelis (changes requested):

alileza:
Hi @italolelis, I believe this is how you expect plugin should looks like, please let me know if there's other thing necessary

Thanks

italolelis (approved):

pkg/plugin/prometheus/setup.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/hellofresh/janus/pkg/router"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	adminRouter router.Router
+)
+
+func init() {
+	plugin.RegisterEventHook(plugin.StartupEvent, onStartup)
```
italolelis: Maybe we should only register these events if Prometheus is enabled? 
I saw that you have a configuration for this but it's only being used to register the route.

pkg/plugin/prometheus/setup.go
```diff
@@ -0,0 +1,61 @@
...
+func onStartup(event interface{}) error {
+	e, ok := event.(plugin.OnStartup)
+	if !ok {
+		return fmt.Errorf("Could not convert event to startup type")
+	}
+	if adminRouter == nil {
+		return errors.New(http.StatusNotFound, "invalid admin router given")
+	}
+	if e.Config.Prometheus.Enabled {
+		adminRouter.GET(e.Config.Prometheus.MetricPath, prometheus.Handler().ServeHTTP)
```
italolelis: What do you think about moving this to `onAdminAPIStartup`? Since `onStartup` and `onAdminAPIStartup` happens concurrently it's hard to guarantee that `adminRouter` will be ready at this point. 
It's more likely that `onStartup` runs before `onAdminAPIStartup`.

italolelis (changes requested):

pkg/plugin/prometheus/setup.go
alileza: my issue with that is i couldn't access the config on `onAdminAPIStartup`, i'm not sure why

pkg/plugin/prometheus/setup.go
alileza: how can i access config from here?

pkg/plugin/prometheus/setup.go
italolelis: True that event doesn't hold the global configuration. I'll work on an improvement to add that to it in the future. But for now, if you catch the configuration `onStartup` you can use it there later. THey are concurrent but `onStartup` starts way before the admin event so very likely it will be ready by then.

alileza:
@italolelis I added new commit that should fix the concurrent issue

alileza:
@italolelis does anything require to be changed from this PR?

vgarvardt:
We're working on adding `prometheus` backend support to `stats-go`, so you'll be able to pull metrics from `janus` pretty soon.

italolelis:
I'm closing this PR in favor of this issue #299

PR #228, opened 2017-11-02
Parse multiple Definitions per api .json file

marcelnubi:
## What does this PR do?
This PR allows file_repository.go to parse an unnamed JSON array of Definition in a single file.


pkg/api/file_repository.go
```diff
@@ -16,6 +16,11 @@ type FileSystemRepository struct {
 	sync.RWMutex
 }
 
+// Type used for  JSON.Unmarshaller
+type definitionList struct {
+	defs []Definition
```
italolelis: I believe this should be a slice of definition pointers? otherwise a lot of coping will happen

pkg/api/file_repository.go
```diff
@@ -36,21 +41,40 @@ func NewFileSystemRepository(dir string) (*FileSystemRepository, error) {
...
+	return nil
+}
+
+func (r *FileSystemRepository) parseDefinition(apiDef []byte) definitionList {
+	appConfigs := definitionList{}
+
+	//Try unmarshalling as if json is an unnamed Array of multiple definitions
+	if err := json.Unmarshal(apiDef, &appConfigs); err != nil {
+		//Try unmarshalling as if json is a single Definition
+		appConfigs.defs = append(appConfigs.defs, *NewDefinition())
```
italolelis: Just so I understand a bit more, why do you do a dereference here? Instead of just using it as a pointer?

italolelis (changes requested):

italolelis:
Hi @marcelnubi,

Thank you for this great contribution. Just one thing, would you mind adding a small explanation about supporting multiple definitions in a single file on the docs (probably [here](docs/quick_start/file_system.md) )?

Thanks again

pkg/api/file_repository.go
marcelnubi: Hello, Italo.
appConfigs.defs is a []Definition and is uninitialized by that point.
It boils down to the Unmarshaller implementation (d *Definition).UnmarshalJSON receiver. This method is called whenever we try to parse a Definition(single or multiple per file) and it points to the first item of the field apiConfigs.defs in case of Single or the N-th in case of Multiple Defition. Actually, this first element could be created with default values Definition{} as a placeholder only since the Unmarshaller will call NewDefinition() anyway.

marcelnubi:
Sure, I'll update the docs accordingly

Edit: I'll write something here as well /docs/quick_start/add_endpoint.md

pkg/api/file_repository.go
italolelis: I got that part. I was just curious why you do 
```
type definitionList struct {
	defs []Definition
}
```

instead of 

```
type definitionList struct {
	defs []*Definition
}
```

But I think I understand now

marcelnubi: Oh, I see. 
So, it's beacause of the Unmarshaller receiver for an array of Definitions: It need to be a slice of Definition(not *Definition)

Thanks

pkg/api/api.go
```diff
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"encoding/json"
```
rafaeljesus, line 4: Missing go format in the imports, actually just add an empty line to separate standard/non-standard pkgs
```go
import (
  "encoding/json"

  "github.com/asaskevich/govalidator"
  "github.com/hellofresh/janus/pkg/proxy"
)
```

pkg/api/api.go
```diff
@@ -45,3 +46,17 @@ func NewDefinition() *Definition {
 func (d *Definition) Validate() (bool, error) {
 	return govalidator.ValidateStruct(d)
 }
+
+// UnmarshalJSON api.Definition JSON.Unmarshaller Implementation
```
rafaeljesus: Implementation could be lowered `implementation`

pkg/api/api.go
```diff
@@ -45,3 +46,17 @@ func NewDefinition() *Definition {
...
+func (d *Definition) UnmarshalJSON(b []byte) error {
+	//Aliasing Definition to avoid recursive call of this method
+	type DefinitionAlias Definition
```
rafaeljesus: Have you considered declaring like this `type definitionAlias Definition` given it is not an exported type?

pkg/api/api.go
```diff
@@ -45,3 +46,17 @@ func NewDefinition() *Definition {
 func (d *Definition) Validate() (bool, error) {
 	return govalidator.ValidateStruct(d)
 }
+
+// UnmarshalJSON api.Definition JSON.Unmarshaller Implementation
+func (d *Definition) UnmarshalJSON(b []byte) error {
+	//Aliasing Definition to avoid recursive call of this method
```
rafaeljesus: Missing some space `// Aliasing`

pkg/api/file_repository.go
```diff
@@ -16,6 +16,11 @@ type FileSystemRepository struct {
 	sync.RWMutex
 }
 
+// Type used for  JSON.Unmarshaller
```
rafaeljesus: Extra space  `for JSON.Unmarshaller`

pkg/api/file_repository.go
```diff
@@ -36,21 +41,40 @@ func NewFileSystemRepository(dir string) (*FileSystemRepository, error) {
 			}
 
 			definition := repo.parseDefinition(appConfigBody)
-			if err = repo.Add(definition); err != nil {
-				log.WithError(err).Error("Can't add the definition to the repository")
-				return nil, err
+			for _, v := range definition.defs {
+				if err = repo.Add(&v); err != nil {
+					log.WithError(err).Error("Can't add the definition" + v.Name + " to the repository")
```
rafaeljesus: Let's use fmt FTW `fmt.Sprintf("Can't add the definition %s to the repository", v.Name)`

pkg/api/file_repository.go
```diff
@@ -36,21 +41,40 @@ func NewFileSystemRepository(dir string) (*FileSystemRepository, error) {
...
 		}
 	}
 
 	return repo, nil
 }
 
-func (r *FileSystemRepository) parseDefinition(apiDef []byte) *Definition {
-	appConfig := NewDefinition()
-	if err := json.Unmarshal(apiDef, appConfig); err != nil {
+//Unmarshal a single Definition
```
rafaeljesus: Missing space `// Unmarshal`

rafaeljesus: TBH I think this method should be inside of `parseDefinition` 
```go
func (r *FileSystemRepository) parseDefinition(apiDef []byte) definitionList {
	appConfigs := definitionList{}

	// try unmarshalling when json is an unnamed Array of multiple definitions
	if err := json.Unmarshal(apiDef, &appConfigs); err != nil {
		// try unmarshalling when json is a single Definition
		appConfigs.defs = append(appConfigs.defs, *NewDefinition())
                if err := json.Unmarshal(apiDef, &appConfigs.defs[0]); err != nil {
		    log.WithError(err).Error("[RPC] --> Couldn't unmarshal api configuration")
	        }
	}

	return appConfigs
}
```

rafaeljesus (changes requested):
@marcelnubi Thanks for your contribution 🎉 

I left some minor comments about coding style

pkg/api/api.go
marcelnubi: That makes sense. I'll fix it, thanks.

pkg/api/file_repository.go
marcelnubi: Hi, @rafaeljesus . Which method should be inside of parseDefinition?

pkg/api/file_repository.go
rafaeljesus: Ah actually we can do even better
```go
log.WithField("name", v.Name).WithError(err).Error("Failed during add definition to the repository")
```

pkg/api/file_repository.go
rafaeljesus: @marcelnubi I was suggesting to get rid of `unmarshalDefinition` method and move it's contents to `parseDefinition` WDYT?

pkg/api/file_repository.go
marcelnubi: Nice, I'll change it as soon as I can

pkg/api/file_repository.go
marcelnubi: I think its right move. Actually, I'm not sure  why I wrote it like that in the first place :P

marcelnubi:
@italolelis & @rafaeljesus , any changes left? I could work on them on this weekend

pkg/api/file_repository.go
```diff
@@ -36,21 +41,33 @@ func NewFileSystemRepository(dir string) (*FileSystemRepository, error) {
...
 }
 
-func (r *FileSystemRepository) parseDefinition(apiDef []byte) *Definition {
-	appConfig := NewDefinition()
-	if err := json.Unmarshal(apiDef, appConfig); err != nil {
-		log.WithError(err).Error("[RPC] --> Couldn't unmarshal api configuration")
+func (r *FileSystemRepository) parseDefinition(apiDef []byte) definitionList {
+	appConfigs := definitionList{}
+
+	//Try unmarshalling as if json is an unnamed Array of multiple definitions
```
rafaeljesus: missing space `// Try`

pkg/api/file_repository.go
```diff
@@ -36,21 +41,33 @@ func NewFileSystemRepository(dir string) (*FileSystemRepository, error) {
...
+	if err := json.Unmarshal(apiDef, &appConfigs); err != nil {
+		//Try unmarshalling as if json is a single Definition
```
rafaeljesus: missing space here too `// Try ...`

marcelnubi: Done

italolelis (approved):

rafaeljesus (approved):

italolelis:
👍 

[Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/228/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway) [Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/228/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway) [Approved with PullApprove](https://pullapprove.com/hellofresh/api-gateway/pull-request/228/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=hellofresh/api-gateway)

aniskhan001:
Hey, I was trying to add multiple endpoints in a single file.

Here are my config:

```
[
    {
        "name": "get",
        "active": true,
        "proxy": {
            "listen_path": "/v1/",
            "upstreams": {
                "balancing": "roundrobin",
                "targets": [
                    { "target": "http://192.168.0.82:3000" }
                ]
            },
            "methods": ["GET"]
        }
    },

    {
        "name": "get-user-promo-id",
        "active": true,
        "proxy": {
            "listen_path": "/v1/promo/{id}",
            "upstreams": {
                "balancing": "roundrobin",
                "targets": [
                    { "target": "http://192.168.0.82:3000/userpromo/{id}" }
                ]
            },
            "methods": ["GET"]
        }
    },

    {
        "name": "get-company-promo",
        "active": true,
        "proxy": {
            "listen_path": "/v1/promo/",
            "upstreams": {
                "balancing": "roundrobin",
                "targets": [
                    { "target": "http://192.168.0.82:3000/promo" }
                ]
            },
            "methods": ["GET"]
        }
    }
]
```

I found that it works only for the last defined endpoint, in this case "get-company-promo". First two don't work. I tried to change the arrangement of them. But it always works for the last one.

Any ideas?

italolelis:
@aniskhan001 Can u create this as a new issue, please? It will help with searching

aniskhan001:
Done at #266

PR #242, opened 2017-11-21
Balancer not working

rafaeljesus:
Fix load balancer not working

### DIFF
--- a/pkg/proxy/register.go
+++ b/pkg/proxy/register.go
@@ -42,10 +42,22 @@ func (p *Register) AddMany(routes []*Route) error {
 func (p *Register) Add(route *Route) error {
 	definition := route.Proxy
 
+	var balancer Balancer
+	if definition.IsBalancerDefined() {
+		log.WithField("balancing_alg", definition.Upstreams.Balancing).Debug("Using a load balancing algorithm")
+
+		var err error
+		balancer, err = NewBalancer(definition.Upstreams.Balancing)
+		if err != nil {
+			log.WithError(err).Error("Could not create a balancer")
+			return err
+		}
+	}
+
 	p.params.Outbound = route.Outbound
 	p.params.InsecureSkipVerify = definition.InsecureSkipVerify
 	handler := &httputil.ReverseProxy{
-		Director:  p.createDirector(definition),
+		Director:  p.createDirector(definition, balancer),
 		Transport: NewTransportWithParams(p.params),
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/proxy/register.go
```diff
@@ -42,10 +42,22 @@ func (p *Register) AddMany(routes []*Route) error {
...
 
+	var balancer Balancer
+	if definition.IsBalancerDefined() {
+		log.WithField("balancing_alg", definition.Upstreams.Balancing).Debug("Using a load balancing algorithm")
+
+		var err error
+		balancer, err = NewBalancer(definition.Upstreams.Balancing)
+		if err != nil {
+			log.WithError(err).Error("Could not create a balancer")
+			return err
```
italolelis: I would do `errors.Wrap(err, "Could not create a balancer")`

pkg/proxy/register.go
```diff
@@ -58,45 +70,33 @@ func (p *Register) Add(route *Route) error {
...
 		// but currently we have several points of definition creation
-		if proxyDefinition.Upstreams != nil && proxyDefinition.Upstreams.Targets != nil && len(proxyDefinition.Upstreams.Targets) > 0 {
-			log.WithField("balancing_alg", proxyDefinition.Upstreams.Balancing).Debug("Using a load balancing algorithm")
-			balancer, err := NewBalancer(proxyDefinition.Upstreams.Balancing)
-			if err != nil {
-				log.WithError(err).Error("Could not create a balancer")
-				return
-			}
-
+		if proxyDefinition.IsBalancerDefined() {
```
italolelis: I would just check for `if balancer != nil {` since you set the balancer before

pkg/proxy/register.go
```diff
@@ -58,45 +70,33 @@ func (p *Register) Add(route *Route) error {
...
 }
 
-func (p *Register) createDirector(proxyDefinition *Definition) func(req *http.Request) {
+func (p *Register) createDirector(proxyDefinition *Definition, balancer Balancer) func(req *http.Request) {
 	return func(req *http.Request) {
-		var target *url.URL
-		var err error
+		var upstreamURL string
 
 		// TODO: find better solution
```
italolelis: Is this comment still necessary?

italolelis (changes requested):

pkg/proxy/register.go
italolelis: And from `"github.com/pkg/errors"`

vgarvardt:
👍

italolelis (approved):

PR #243, opened 2017-11-23
Check GitHub permissions

rafaeljesus:
Sets `is_admin` into the jwt token when the chosen provider is Github

### DIFF
--- a/pkg/jwt/github/provider.go
+++ b/pkg/jwt/github/provider.go
@@ -53,22 +53,18 @@ func (gp *Provider) GetClaims(httpClient *http.Client) (jwt.MapClaims, error) {
 
 	go func() {
 		defer wg.Done()
-		res, err := client.CurrentUser(httpClient)
+		user, err = client.CurrentUser(httpClient)
 		if err != nil {
-			err = append(errs, errors.Wrap(err, "failed to get github users"))
-			return
+			err = errors.Wrap(err, "failed to get github users")
 		}
-		user = res
 	}()
 
 	go func() {
 		defer wg.Done()
-		res, err := client.Teams(httpClient)
+		usersOrgTeams, err = client.Teams(httpClient)
 		if err != nil {
-			err = append(errs, errors.Wrap(err, "failed to get github teams"))
-			return
+			err = errors.Wrap(err, "failed to get github teams")
 		}
-		usersOrgTeams = res
 	}()
 
 	wg.Wait()
--- a/pkg/jwt/github/provider.go
+++ b/pkg/jwt/github/provider.go
@@ -11,10 +11,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-const (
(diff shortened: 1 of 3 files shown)


pkg/jwt/github/provider.go
```diff
@@ -15,35 +22,70 @@ func init() {
...
 	client := NewClient()
 
-	user, err := client.CurrentUser(httpClient)
-	if err != nil {
-		return nil, err
+	var (
+		wg            sync.WaitGroup
+		user          *github.User
+		usersOrgTeams OrganizationTeams
+		errs          []error
```
italolelis: `err error`

pkg/jwt/github/provider.go
```diff
@@ -15,35 +22,70 @@ func init() {
...
+	)
+
+	wg.Add(2)
+
+	go func() {
+		defer wg.Done()
+		res, err := client.CurrentUser(httpClient)
+		if err != nil {
+			errs = append(errs, errors.Wrap(err, "failed to get github users"))
```
italolelis: `err = errors.Wrap(err, "failed to get github users")`

pkg/jwt/github/provider.go
```diff
@@ -15,35 +22,70 @@ func init() {
...
+			return
+		}
+		user = res
+	}()
+
+	go func() {
+		defer wg.Done()
+		res, err := client.Teams(httpClient)
+		if err != nil {
+			errs = append(errs, errors.Wrap(err, "failed to get github teams"))
```
italolelis: `err = errors.Wrap(err, "failed to get github teams")`

pkg/jwt/github/provider.go
```diff
@@ -15,35 +22,70 @@ func init() {
...
+			return
+		}
+		usersOrgTeams = res
+	}()
+
+	wg.Wait()
+
+	if len(errs) > 0 {
```
italolelis: `if err != nil`

pkg/jwt/github/provider.go
```diff
@@ -2,10 +2,17 @@ package github
...
 
 	"github.com/dgrijalva/jwt-go"
+	"github.com/google/go-github/github"
 	"github.com/hellofresh/janus/pkg/config"
 	"github.com/hellofresh/janus/pkg/jwt/provider"
+	"github.com/pkg/errors"
+)
+
+const (
+	janusOwners = "janus-owners"
```
italolelis: This should come from an environment variable `JANUS_ADMIN_TEAM` and should be passed to the global config

italolelis (changes requested):

pkg/jwt/github/provider.go
rafaeljesus: I think it's a bit harder to understand the code aggregating the result from 2 calls to a single variable, but not a big deal will change

italolelis (approved):

Issue #249, opened 2017-12-07
Valide a plugin struct


rjtch:
i tried to add plugin to my api and also to my oauth server and i only gate a validation error. How can i make it validated. my request below
```
PUT /apis/alain HTTP/1.1
Accept: application/json, */*
Accept-Encoding: gzip, deflate
Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1MTI2NTc0MjMsImlhdCI6MTUxMjY1MzgyM30._NmN-qTzeNjR4pvVdt-c8MwWn-3OMivhokDNi0FQxYU
Connection: keep-alive
Content-Length: 165
Content-Type: application/json
Host: janus:8081
User-Agent: HTTPie/0.9.9

{
    "plugins": [
        {
            "config": {
                "limit": "5-M",
                "policy": "local"
            },
            "enabled": true,
            "name": "rate-limiting"
        }
    ]
}

HTTP/1.1 400 Bad Request
Content-Encoding: gzip
Content-Length: 74
Content-Type: application/json
Date: Thu, 07 Dec 2017 13:58:24 GMT
Vary: Origin

{
    "error": "name: non zero value required;proxy: non zero value required"
}
```

italolelis:
Hey @rjtch,

That's probably happening because you are using the PUT method. According to how HTTP methods should work, when using a PUT you should provide the whole API definition. If you do that you won't get this error. 
We will also work on improving this error message that doesn't have much sense.

rafaeljesus:
@rjtch @hellofresh/janus is going to support `Patch` endpoint pretty soon, stay tuned 😉

PR #250, opened 2017-12-13
Feature/improve validation msg

rafaeljesus:
Improves existing validation error messages

Closes https://github.com/hellofresh/janus/issues/249

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -30,7 +30,7 @@ type Definition struct {
 
 // HealthCheck represents the health check configs
 type HealthCheck struct {
-	URL     string `bson:"url" json:"url" valid:"url~health_check.url is required"`
+	URL     string `bson:"url" json:"url" valid:"url"`
 	Timeout int    `bson:"timeout" json:"timeout"`
 }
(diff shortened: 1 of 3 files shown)


pkg/api/api.go
```diff
@@ -21,9 +21,9 @@ type Plugin struct {
 
 // Definition Represents an API that you want to proxy
 type Definition struct {
-	Name        string            `bson:"name" json:"name" valid:"required"`
+	Name        string            `bson:"name" json:"name" valid:"required~name is required"`
```
vgarvardt, line 24: please add integration tests to check if these errors are really go to a response

vgarvardt:
👍

pkg/api/api.go
rafaeljesus: Added, however adding validation and integration tests to all other fields like `upstreams.targets.target` is out of scope of this PR, cos ATM api allows inserting empty upstreams, it will be fixed here https://github.com/hellofresh/janus/issues/251

italolelis (approved):

PR #258, opened 2018-01-05
Fix possible race condition

rafaeljesus:
Fix possible race condition

### DIFF
--- a/pkg/jwt/github/provider.go
+++ b/pkg/jwt/github/provider.go
@@ -46,31 +46,30 @@ func (gp *Provider) GetClaims(httpClient *http.Client) (jwt.MapClaims, error) {
 		wg            sync.WaitGroup
 		user          *github.User
 		usersOrgTeams OrganizationTeams
-		err           error
+		errs          []error
 	)
 
 	wg.Add(2)
-
-	go func() {
+	go func(user *github.User, client Client, httpClient *http.Client) {
 		defer wg.Done()
-		user, err = client.CurrentUser(httpClient)
+		user, err := client.CurrentUser(httpClient)
 		if err != nil {
-			err = errors.Wrap(err, "failed to get github users")
+			errs = append(errs, errors.Wrap(err, "failed to get github users"))
 		}
-	}()
+	}(user, client, httpClient)
 
-	go func() {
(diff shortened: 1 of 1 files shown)


pkg/jwt/github/provider.go
```diff
@@ -46,31 +46,30 @@ func (gp *Provider) GetClaims(httpClient *http.Client) (jwt.MapClaims, error) {
...
 		}
-	}()
+	}(usersOrgTeams, client, httpClient)
 
 	wg.Wait()
 
-	if err != nil {
-		return nil, err
+	if len(errs) > 0 {
+		return nil, errs[0]
```
italolelis: Why return only the first error? we should return all errors, right?

italolelis (changes requested):

pkg/jwt/github/provider.go
rafaeljesus: I think it's fine to return only one error, the normal flow would be execute one call over another and we would be returning one error though. This is such a tiny optimisation for doing IO don't think it desires any break on return signature

italolelis: I don't mean to change the return signature, that's why I was using `errors.Wrap`, because in the end, it was returning all errors as an error and if you did `err.Error()` you could see the string of all errors or if you use `error.Cause(err)` you could know what error was happening. 
I don't think it's ok to hide the other errors. Also, I don't get it why you changed the error implementation here.

rafaeljesus: Ok, so we can iterate over the errors slice and combine the errors to a single error, something in this line:

```go
if len(errs) > 0 {
  var err error

  for _, e := range errs {
    err = errors.Wrap(err, e.Error())
  }

  return nil, err
}
```

Current we have a race condition while assigning err value among goroutines:

```go
var err error

go func() {
  // let's say CurrentUser cll fails and returns error
  err = client.CurrentUser()
}()

go func() {
  // Now if Teams call succeed it will assign nil to err var
  // err will be nil 
  err = client.Teams()
}()
```

Makes sense?

italolelis: Can't we pass the error to the goroutine? WIthout using it from outside?
Or create an error channel and wait for it one it's done? I think it's more elegant and also will catch all errors

italolelis (approved):

Issue #260, opened 2018-01-17
Format taget URL with parameters captured from the listened path


olvlvl:
Hi,

I'm trying to get the target URL formatted with the parameters captured form the listened path, but the whole path is appended to the target path. Please consider the following spec (excerpt):

```
{
    "proxy" : {
        "listen_path" : "/api/recipes/{id:[\\da-f]{24}}",
        "upstreams" : {
            "targets" : [
                {
                    "target" : "http://localhost:8069/recipes/{id}"
                }
            ]
        }
    },
```

I'd like paths such as `/api/recipes/5981814cfeb6f31e78535e7d` to be redirected to `http://localhost:8069/recipes/5981814cfeb6f31e78535e7d`.

**Expected behavior:**

`/api/recipes/5981814cfeb6f31e78535e7d` is redirected to `http://localhost:8069/recipes/5981814cfeb6f31e78535e7d`.

**Observed behavior:**

`/api/recipes/5981814cfeb6f31e78535e7d` is redirected to `http://localhost:8069/recipes/%7Bid%7D/api/recipes/5981814cfeb6f31e78535e7d`.

**Janus version:** Latest
**OS and version:** Linux

vgarvardt:
Looks like this is what you need:

```
{
    "proxy" : {
        "listen_path" : "/api/recipes/*",
        "append_path": true,
        "upstreams" : {
            "targets" : [
                {
                    "target" : "http://localhost:8069/recipes/"
                }
            ]
        }
    },
```

olvlvl:
Not all endpoints under `/recipes/` can be redirected to the recipe service e.g. `/recipes/{id}/favorites`, `/recipes/{id}/ratings`, `/recipes/{id}/menus`, …

stefanoj3:
I wrote a quick and **-very dirty-** test:
```
pkg/proxy/register_test.go
```

```
func TestT(t *testing.T) {
	d := &Definition{
		ListenPath: "/api/recipes/*/activate",
		Upstreams: &Upstreams{
			Balancing: "roundrobin",
			Targets:   []*Target{{Target: "http://recipes-service:9089/recipes/"}},
		},
		Methods: []string{"ALL"},
		AppendPath: true,
		StripPath: true,
	}

	r := router.NewChiRouter()
	regist := createRegister(r)

	fun := regist.createDirector(d, NewRoundrobinBalancer())

	req, err := http.NewRequest("GET", "http://localhost/api/recipes/1234/activate", nil)
[...]

PR #262, opened 2018-01-24
Support url parameters

stefanoj3:
## What does this PR do?
Adds supports for named parameters in the target url.

**Example**
Given this configuration
```
Definition{
	ListenPath: "/api/recipes/{id:[\\da-f]{24}}",
	Upstreams: &Upstreams{
		Balancing: "roundrobin",
		Targets:   []*Target{{Target: "http://localhost:9089/recipes/{id}"}},
	},
	Methods:    []string{"ALL"},
	StripPath:  false,
	AppendPath: false,
}
```

If I call `/api/recipes/5252b1b5301bbf46038b473f` my call should be forwarded to `http://localhost:9089/recipes/5252b1b5301bbf46038b473f`

### DIFF
--- a/features/Proxy.feature
+++ b/features/Proxy.feature
@@ -207,3 +207,75 @@ Feature: Proxy requests to upstream.
 
         When I request "/posts-public/nested/path" path with "GET" method
         Then I should receive 404 response code
+
+        Given request JSON payload:
+            """
+            {
+                "name":"one parameter",
+                "active":true,
+                "proxy":{
+                    "preserve_host":false,
+                    "listen_path":"/api/recipes/{id:[\\da-f]{24}}",
+                    "upstreams":{
+                        "balancing":"roundrobin",
+                        "targets":[
+                            {
+                                "target":"http://localhost:9089/recipes/{id}"
+                            }
+                        ]
+                    },
+                    "strip_path":false,
+                    "append_path":false,
+                    "enable_load_balancing":false,
+                    "methods":[
+                        "GET"
+                    ]
+                },
+                "health_check":{
+                    "url":"https://example.com/status"
+                }
+            }
+            """
+        When I request "/apis" API path with "POST" method
+        Then I should receive 201 response code
+
+        When I wait for a while
+        And I request "/api/recipes/5252b1b5301bbf46038b473f" path with "GET" method
+        Then I should receive 200 response code
+        And the response should contain "I'm a slug"
+
+        Given request JSON payload:
+            """
+            {
+                "name":"two parameters",
+                "active":true,
+                "proxy":{
+                    "preserve_host":false,
+                    "listen_path":"/api/recipes/{recipeId:[\\da-f]{24}}/menus/{menuId:[\\da-f]{24}}",
+                    "upstreams":{
+                        "balancing":"roundrobin",
+                        "targets":[
+                            {
+                                "target":"http://localhost:9089/recipes/{recipeId}/menus/{menuId}"
+                            }
+                        ]
+                    },
+                    "strip_path":false,
+                    "append_path":false,
+                    "enable_load_balancing":false,
+                    "methods":[
+                        "GET"
+                    ]
+                },
+                "health_check":{
+                    "url":"https://example.com/status"
+                }
+            }
+            """
+        When I request "/apis" API path with "POST" method
+        Then I should receive 201 response code
+
+        When I wait for a while
+        And I request "/api/recipes/5252b1b5301bbf46038b473f/menus/6362b1b5301bbf46038b4766" path with "GET" method
+        Then I should receive 200 response code
+        And the response should contain "A menu description"
--- a/pkg/proxy/register.go
+++ b/pkg/proxy/register.go
@@ -1,6 +1,7 @@
 package proxy
 
 import (
+	"fmt"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
@@ -73,6 +74,9 @@ func (p *Register) Add(route *Route) error {
 }
 
 func (p *Register) createDirector(proxyDefinition *Definition, balancer Balancer) func(req *http.Request) {
+	paramNameExtractor := router.NewListenPathParamNameExtractor()
+	matcher := router.NewListenPathMatcher()
+
 	return func(req *http.Request) {
 		var upstreamURL string
 		if proxyDefinition.IsBalancerDefined() && balancer != nil {
@@ -106,7 +110,6 @@ func (p *Register) createDirector(proxyDefinition *Definition, balancer Balancer
 
 		if proxyDefinition.StripPath {
 			path = singleJoiningSlash(target.Path, req.URL.Path)
-			matcher := router.NewListenPathMatcher()
 			listenPath := matcher.Extract(proxyDefinition.ListenPath)
 
 			log.WithField("listen_path", listenPath).Debug("Stripping listen path")
@@ -114,6 +117,14 @@ func (p *Register) createDirector(proxyDefinition *Definition, balancer Balancer
 			if !strings.HasSuffix(target.Path, "/") && strings.HasSuffix(path, "/") {
 				path = path[:len(path)-1]
 			}
+		}
+
+		paramNames := paramNameExtractor.Extract(path)
+		parametrizedPath, err := p.applyParameters(req, path, paramNames)
+		if err != nil {
+			log.WithError(err).Warn("Unable to extract param from request")
+		} else {
+			path = parametrizedPath
 		}
 
 		log.WithField("path", path).Debug("Upstream Path")
@@ -150,6 +161,25 @@ func (p *Register) doRegister(listenPath string, handler http.HandlerFunc, metho
 			}
 		}
 	}
+}
+
+func (p *Register) applyParameters(req *http.Request, path string, paramNames []string) (string, error) {
+	for _, paramName := range paramNames {
+		paramValue := router.URLParam(req, paramName)
+
+		if len(paramValue) == 0 {
+			return "", errors.Errorf("unable to extract {%s} from request", paramName)
+		}
+
+		path = strings.Replace(
+			path,
+			fmt.Sprintf("{%s}", paramName),
+			paramValue,
+			-1,
+		)
+	}
+
+	return path, nil
 }
 
 func cleanSlashes(a string) string {
--- a/pkg/proxy/register_test.go
+++ b/pkg/proxy/register_test.go
@@ -48,23 +48,39 @@ var tests = []struct {
 		expectedContentType: "application/json; charset=utf-8",
 		expectedCode:        http.StatusOK,
 	},
+	{
+		description:         "Get one recipe - parameter interpolation",
+		url:                 "/api/recipes/5252b1b5301bbf46038b473f",
+		expectedContentType: "application/json; charset=utf-8",
+		expectedCode:        http.StatusOK,
+	},
+	{
+		description:         "No parameter to interpolate",
+		url:                 "/api/recipes/search",
+		expectedContentType: "application/json",
+		expectedCode:        http.StatusNotFound,
+	},
 }
 
 func TestSuccessfulProxy(t *testing.T) {
+	t.Parallel()
+
 	log.SetOutput(ioutil.Discard)
 
 	ts := test.NewServer(createRegisterAndRouter())
 	defer ts.Close()
 
 	for _, tc := range tests {
-		res, err := ts.Do(tc.method, tc.url, make(map[string]string))
-		assert.NoError(t, err)
-		if res != nil {
-			defer res.Body.Close()
-		}
+		t.Run(tc.description, func(t *testing.T) {
+			res, err := ts.Do(tc.method, tc.url, make(map[string]string))
+			assert.NoError(t, err)
+			if res != nil {
+				defer res.Body.Close()
+			}
 
-		assert.Equal(t, tc.expectedContentType, res.Header.Get("Content-Type"), tc.description)
-		assert.Equal(t, tc.expectedCode, res.StatusCode, tc.description)
+			assert.Equal(t, tc.expectedContentType, res.Header.Get("Content-Type"), tc.description)
+			assert.Equal(t, tc.expectedCode, res.StatusCode, tc.description)
+		})
 	}
 }
 
@@ -93,6 +109,22 @@ func createProxyDefinitions() []*Definition {
 			AppendPath:  true,
 			Upstreams:   &Upstreams{},
 			Methods:     []string{"GET"},
+		},
+		{
+			ListenPath: "/api/recipes/{id:[\\da-f]{24}}",
+			Upstreams: &Upstreams{
+				Balancing: "roundrobin",
+				Targets:   []*Target{{Target: "http://localhost:9089/recipes/{id}"}},
+			},
+			Methods: []string{"GET"},
+		},
+		{
+			ListenPath: "/api/recipes/search",
+			Upstreams: &Upstreams{
+				Balancing: "roundrobin",
+				Targets:   []*Target{{Target: "http://localhost:9089/recipes/{id}"}},
+			},
+			Methods: []string{"GET"},
 		},
 	}
 }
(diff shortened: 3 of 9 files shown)


olvlvl:
WANT! https://github.com/hellofresh/janus/issues/260

pkg/router/listen_path_parameter_name_extractor.go
```diff
@@ -0,0 +1,30 @@
+package router
+
+import "regexp"
+
+const (
+	parameterMatchRule = `\{([a-z\d]+?)\}`
```
olvlvl: How about `/\{([^/}]+)\}/`?

pkg/proxy/register.go
```diff
@@ -116,6 +119,22 @@ func (p *Register) createDirector(proxyDefinition *Definition, balancer Balancer
 			}
 		}
 
+		parameterNames := parameterNameExtractor.Extract(path)
+		for _, parameterName := range parameterNames {
+			parameterValue := router.URLParam(req, parameterName)
+
+			if len(parameterValue) == 0 {
+				continue
```
olvlvl: I'm not sure about this one, if `id` is an empty string then the path will have `{id}` in it.

stefanoj3: I was in doubt as well on this, on a second look I should probably log the error to give visibility to it, but the behaviour should stay the same, the role of this component is to alter the request to proxy it somewhere else, but we cannot give an actual response from here

In the same file the approach is the same: log and return: pkg/proxy/register.go:96

What do you think?

pkg/router/listen_path_parameter_name_extractor.go
stefanoj3: I need the group to extract the parameter inside the braches

olvlvl: The group is in mine as well :)

stefanoj3: Now it is, I saw you 😛

pkg/router/listen_path_parameter_name_extractor.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// NewListenPathParameterNameExtractor creates a new instance ListenPathParameterNameExtractor
+func NewListenPathParameterNameExtractor() *ListenPathParameterNameExtractor {
+	return &ListenPathParameterNameExtractor{regexp.MustCompile(parameterMatchRule)}
+}
+
+// Extract takes the usable part of the listen path an extracts parameter names
+func (l *ListenPathParameterNameExtractor) Extract(listenPath string) []string {
+	submatches := l.reg.FindAllStringSubmatch(listenPath, -1)
```
olvlvl: What do you think about looking of `{` first and return an empty array if it's not there? Do you think it could speed up things?

stefanoj3: Performing on 10000 urls the difference is of 12ms on the total execution time, not sure it's worth adding a bit of complexity there, the gain seems pretty small

~64ms vs ~76ms

pkg/proxy/register.go
```diff
@@ -73,6 +74,9 @@ func (p *Register) Add(route *Route) error {
 }
 
 func (p *Register) createDirector(proxyDefinition *Definition, balancer Balancer) func(req *http.Request) {
+	parameterNameExtractor := router.NewListenPathParameterNameExtractor()
```
rafaeljesus: I think we could reduce the name a bit, smth like `paramNameExtractor 
 := NewListenPathParamNameExtractor`

pkg/proxy/register.go
```diff
@@ -116,6 +119,29 @@ func (p *Register) createDirector(proxyDefinition *Definition, balancer Balancer
 			}
 		}
 
+		parameterNames := parameterNameExtractor.Extract(path)
+		for _, parameterName := range parameterNames {
+			parameterValue := router.URLParam(req, parameterName)
+
+			if len(parameterValue) == 0 {
+				log.WithField(
```
rafaeljesus: Here we could:
```go
log.WithFields(log.Fields{
  "paramName": paramName,
  "path": path,
}).Warn("Unable to extract parameter from request")
```

I'm not really sure we should log error and continue with the request, so perhaps warn is more appropriate

pkg/proxy/register.go
```diff
@@ -116,6 +119,29 @@ func (p *Register) createDirector(proxyDefinition *Definition, balancer Balancer
 			}
 		}
 
+		parameterNames := parameterNameExtractor.Extract(path)
+		for _, parameterName := range parameterNames {
+			parameterValue := router.URLParam(req, parameterName)
```
rafaeljesus: I would rename all parameter o param

pkg/router/listen_path_parameter_name_extractor.go
```diff
@@ -0,0 +1,30 @@
...
+
+// NewListenPathParameterNameExtractor creates a new instance ListenPathParameterNameExtractor
+func NewListenPathParameterNameExtractor() *ListenPathParameterNameExtractor {
+	return &ListenPathParameterNameExtractor{regexp.MustCompile(parameterMatchRule)}
+}
+
+// Extract takes the usable part of the listen path and extracts parameter names
+func (l *ListenPathParameterNameExtractor) Extract(listenPath string) []string {
+	submatches := l.reg.FindAllStringSubmatch(listenPath, -1)
+
```
rafaeljesus: we could remove this empty line

pkg/proxy/register_test.go
```diff
@@ -117,3 +117,76 @@ func createRegister(r router.Router) *Register {
...
+	log.SetOutput(ioutil.Discard)
+
+	def := &Definition{
+		ListenPath: "/api/recipes/{id:[\\da-f]{24}}",
+		Upstreams: &Upstreams{
+			Balancing: "roundrobin",
+			Targets:   []*Target{{Target: "http://localhost:9089/recipes/{id}"}},
+		},
+		Methods:    []string{"ALL"},
+		StripPath:  false,
```
rafaeljesus: we could omit `StripPath` and `AppendPath` given they are false

pkg/proxy/register_test.go
```diff
@@ -117,3 +117,76 @@ func createRegister(r router.Router) *Register {
...
+		AppendPath: false,
+	}
+	route := NewRoute(def)
+
+	chiRouter := router.NewChiRouter()
+
+	statsClient, _ := stats.NewClient("noop://", "")
```
rafaeljesus: Always catch errors, for testing I'd recommend `require.NoError(t, err)`, just remember to import `github.com/stretchr/testify/require`

pkg/proxy/register_test.go
```diff
@@ -117,3 +117,76 @@ func createRegister(r router.Router) *Register {
...
+	register := NewRegister(chiRouter, Params{StatsClient: statsClient})
+	register.Add(route)
+
+	ts := test.NewServer(chiRouter)
+	defer ts.Close()
+
+	res, err := ts.Do("GET", "/api/recipes/5252b1b5301bbf46038b473f", make(map[string]string))
+	if err != nil {
+		t.Fatal("Call failed", err.Error())
```
rafaeljesus: `require.NoError(t, err)` here too

pkg/proxy/register_test.go
vgarvardt: @rafaeljesus not necessary in this case

pkg/proxy/register_test.go
```diff
@@ -117,3 +117,76 @@ func createRegister(r router.Router) *Register {
...
+	}
+
+	defer res.Body.Close()
+
+	assert.Equal(t, 200, res.StatusCode)
+
+	body, _ := ioutil.ReadAll(res.Body)
```
rafaeljesus: Here too `require.NoError(t, err)`

pkg/proxy/register_test.go
```diff
@@ -117,3 +117,76 @@ func createRegister(r router.Router) *Register {
...
+	log.SetOutput(ioutil.Discard)
+
+	def := &Definition{
+		ListenPath: "/api/recipes/search",
+		Upstreams: &Upstreams{
+			Balancing: "roundrobin",
+			Targets:   []*Target{{Target: "http://localhost:9089/recipes/{id}"}},
+		},
+		Methods:    []string{"ALL"},
+		StripPath:  false,
```
rafaeljesus: We can omit those false vars

pkg/proxy/register_test.go
```diff
@@ -117,3 +117,76 @@ func createRegister(r router.Router) *Register {
...
+		AppendPath: false,
+	}
+	route := NewRoute(def)
+
+	chiRouter := router.NewChiRouter()
+
+	statsClient, _ := stats.NewClient("noop://", "")
```
rafaeljesus: Same regarding errors

pkg/proxy/register_test.go
```diff
@@ -117,3 +117,76 @@ func createRegister(r router.Router) *Register {
...
+	register := NewRegister(chiRouter, Params{StatsClient: statsClient})
+	register.Add(route)
+
+	ts := test.NewServer(chiRouter)
+	defer ts.Close()
+
+	res, err := ts.Do("GET", "/api/recipes/search", make(map[string]string))
+	if err != nil {
+		t.Fatal("Call failed", err.Error())
```
rafaeljesus: `require.NoError(t, err)` too

pkg/proxy/register_test.go
vgarvardt: no need to check for error in tests for `noop` backend

stefanoj3:
@rafaeljesus updated with the changes you requested

pkg/proxy/register.go
```diff
@@ -116,6 +119,26 @@ func (p *Register) createDirector(proxyDefinition *Definition, balancer Balancer
 			}
 		}
 
+		paramNames := paramNameExtractor.Extract(path)
+		for _, paramName := range paramNames {
```
italolelis: This should probably be encapsulated into a function. Maybe something like `applyParameters` or anything on the same line of thinking. I don't know where this function should go, maybe the `paramNameExtractor`?

stefanoj3: Then I will put it in a function in this same file, I would like to leave the `paramNameExtractor` unaware of request, context and so on.
We already have functions playing with the path in here like: 
```
cleanSlashes @ pkg/proxy/register.go:178 
singleJoiningSlash @ pkg/proxy/register.go:193
```

Sounds good?

pkg/proxy/register_test.go
```diff
@@ -117,3 +118,70 @@ func createRegister(r router.Router) *Register {
 
 	return register
 }
+
+func TestParametersReplacement(t *testing.T) {
+	log.SetOutput(ioutil.Discard)
+
+	def := &Definition{
```
rafaeljesus: Couldn't this be inside of createProxyDefinitions func?

pkg/proxy/register.go
italolelis: Yep that's ok

pkg/proxy/register_test.go
stefanoj3: I wanted to have it isolated to reduce as much as possible any external factor influencing the test, do you want me to move it in?

rafaeljesus: Also I would suggest such structure for this tests if possible:

```go
func TestSuccessfulProxy(t *testing.T) {
    t.Parallel()

    log.SetOutput(ioutil.Discard)

    tests := []struct {
        scenario, url, expectedContentType string
        expectedCode                       int
        function                           func(*testing.T, *test.Server, string, string, string, int)
    }{
        {
            scenario:            "Get example route",
            url:                 "/example",
            expectedContentType: "application/json; charset=utf-8",
            expectedCode:        http.StatusOK,
            function:            testGetExampleRoute,
        },
        {
            scenario:            "Get invalid route",
            url:                 "/invalid-route",
            expectedContentType: "text/plain; charset=utf-8",
            expectedCode:        http.StatusNotFound,
            function:            testGetInvalidRoute,
        },
    }

    ts := test.NewServer(createRegisterAndRouter())
    defer ts.Close()

    for _, test := range tests {
        t.Run(test.scenario, func(t *testing.T) {
            test.function(t, ts, test.url, test.expectedContentType, test.scenario, test.expectedCode)
        })
    }
}

func testGetExampleRoute(t *testing.T, ts *test.Server, url, contentType, description string, code int) {
    res, err := ts.Do(http.MethodGet, url, make(map[string]string))
    assert.NoError(t, err)
    if res != nil {
        defer res.Body.Close()
    }

    assert.Equal(t, contentType, res.Header.Get("Content-Type"), description)
    assert.Equal(t, code, res.StatusCode, description)
}

... more tests here
```

rafaeljesus: If they do the same thing better to move them, when reading the code one can assume there's something different with these tests

stefanoj3:
Updated with the requested changes and with the help of @rafaeljesus I added also an automation test

features/Proxy.feature
```diff
@@ -207,3 +207,40 @@ Feature: Proxy requests to upstream.
...
         Then I should receive 404 response code
+
+        Given request JSON payload:
+            """
+            {
+                "name":"parameters_interpolation",
+                "active":true,
+                "proxy":{
+                    "preserve_host":false,
+                    "listen_path":"/api/recipes/{id:[\\da-f]{24}}",
```
rafaeljesus: Could we also test `/api/recipes/{id:[\\da-f]{24}}/menus/{id:[\\da-f]{24}}` to ensure it also works with more than one parameter?

features/Proxy.feature
```diff
@@ -207,3 +207,40 @@ Feature: Proxy requests to upstream.
...
+                    "upstreams":{
+                        "balancing":"roundrobin",
+                        "targets":[
+                            {
+                                "target":"http://localhost:9089/recipes/{id}"
```
rafaeljesus, line 223: https://github.com/hellofresh/janus/pull/262/files#r163896506 I mean here `http://localhost:9089/recipes/{id}"/menus/{menuID}`

features/Proxy.feature
stefanoj3: sure, I'll add it, it should be then
`/api/recipes/{recipeId:[\\da-f]{24}}/menus/{menuId:[\\da-f]{24}}`

pkg/proxy/register.go
```diff
@@ -116,6 +119,14 @@ func (p *Register) createDirector(proxyDefinition *Definition, balancer Balancer
 			}
 		}
 
+		paramNames := paramNameExtractor.Extract(path)
+		parametrizedPath, err := applyParameters(path, paramNames, req)
```
rafaeljesus: 👍

pkg/proxy/register.go
```diff
@@ -116,6 +119,14 @@ func (p *Register) createDirector(proxyDefinition *Definition, balancer Balancer
...
+		if err != nil {
+			log.WithField("err", err.Error()).Warn("Unable to extract param from request")
```
rafaeljesus: here we could `log.WithError(err).Warn("Unable to extract param from request")`

pkg/proxy/register.go
rafaeljesus: here we should have `h.applyParameters` for private funcs

rafaeljesus: I mean let's make it private rather than global

rafaeljesus: Ah and if possible invert the order of params to have the stronger dependencies first e.g `h.applyParameters(req, path, paramNames)`.

rafaeljesus:
@stefanoj3 we could add this new feature on the docs as well https://github.com/hellofresh/janus/docs/proxy/routing_capabilities.md a simple example is enough

stefanoj3:
Doc updated

rafaeljesus (approved):

rafaeljesus (approved):

italolelis (approved):

rafaeljesus:
@stefanoj3 @olvlvl really appreciate your effort to push this feature, thank you

olvlvl:
Thanks guys! I'm really glad about this feature!

PR #263, opened 2018-01-25
Distributed Tracing: Implementing Jaeger

italolelis:
## What does this PR do?

- Implements Jaeger as a new backend for tracing
- Improves our Dockerfile to not depend on GitHub
- Removes Appdash from the supported tracing systems


cmd/janus/version.go
```diff
@@ -1,10 +1,25 @@
...
+		Aliases: []string{"v"},
+		Run: func(cmd *cobra.Command, args []string) {
+			RunVersion()
+		},
+	}
+}
+
+// RunVersion runs the command to print the current version
+func RunVersion() {
+	log.Infof("janus %s", version)
```
vgarvardt: log may be set to write to the output different than stdout/strerr, then user will not see anything, `fmt.Printf(...)` is better in this case IMO

Gopkg.toml
```diff
@@ -67,9 +68,9 @@
   name = "github.com/kelseyhightower/envconfig"
   version = "1.3.0"
 
-[[constraint]]
-  branch = "master"
+[[constraint]] 
   name = "github.com/mitchellh/mapstructure"
+  branch = "master"
```
rafaeljesus: what about pinning a version? if I am not mistaken, we had trouble one with mapstructure not being versioned

italolelis: They don't use semver, which makes this difficult, we will have to pin to a review (which is not good), but we could do it

pkg/opentracing/opentracing.go
```diff
@@ -66,3 +44,41 @@ func FromContext(ctx context.Context, name string) opentracing.Span {
 func ToContext(r *http.Request, span opentracing.Span) *http.Request {
 	return r.WithContext(opentracing.ContextWithSpan(r.Context(), span))
 }
+
+func buildGCloud(config config.GoogleCloudTracing) (opentracing.Tracer, error) {
+	tracer, err := gcloudtracer.NewTracer(
```
rafaeljesus: we could just
```go
return gcloudtracer.NewTracer(
...
)
```

pkg/opentracing/opentracing.go
```diff
@@ -66,3 +44,41 @@ func FromContext(ctx context.Context, name string) opentracing.Span {
...
+}
+
+func buildJaeger(c config.JaegerTracing) (opentracing.Tracer, io.Closer, error) {
+	cfg := jaegercfg.Configuration{
+		Sampler: &jaegercfg.SamplerConfig{
+			Type:  jaeger.SamplerTypeConst,
+			Param: 1,
+		},
+		Reporter: &jaegercfg.ReporterConfig{
+			LogSpans:            true,
```
rafaeljesus: Should `LogSpans` and `BufferFlushInterval` be configurable?

pkg/opentracing/opentracing.go
```diff
@@ -2,57 +2,35 @@ package opentracing
...
-		} else {
-			appdashFields.Debug("Using remote appdash server as tracing system")
-		}
-
-		return server.GetTracer(), nil
-	} else {
+// Build a tracer based on the configuration provided
+func Build(config config.Tracing) (opentracing.Tracer, io.Closer, error) {
+	switch config.Tracer {
+	case "gcloud":
```
rafaeljesus: we could extract a const for these strings

pkg/opentracing/opentracing.go
```diff
@@ -66,3 +45,42 @@ func FromContext(ctx context.Context, name string) opentracing.Span {
...
+	cfg := jaegercfg.Configuration{
+		Sampler: &jaegercfg.SamplerConfig{
+			Type:  jaeger.SamplerTypeConst,
+			Param: 1,
+		},
+		Reporter: &jaegercfg.ReporterConfig{
+			LogSpans:            c.LogSpans,
+			BufferFlushInterval: bufferFLushInterval,
+			LocalAgentHostPort:  c.DSN,
+			QueueSize:           c.QueueSize,
```
rafaeljesus, line 82: 👍

rafaeljesus (approved):

rafaeljesus (approved):

rafaeljesus:
:shipit:

rafaeljesus (approved):

PR #269, opened 2018-02-27
PTII-145 Feature circuit breaker

c4pone:
## What does this PR do?
This will enable circuit breakers for each proxy


cmd/janus/init.go
```diff
@@ -58,6 +63,26 @@ func initDistributedTracing() io.Closer {
...
+	metricCollector.Registry.Register(c.NewStatsdCollector)
+	if err != nil {
+		log.WithError(err).Error("Error initializing statsd client for circuit breaker")
+	}
+
+	// Setup hystrix dashboard stream
+	if globalConfig.CircuitBreaker.DashboardEnabled {
+		hystrixStreamHandler := hystrix.NewStreamHandler()
+		hystrixStreamHandler.Start()
+		go http.ListenAndServe(net.JoinHostPort("", "81"), hystrixStreamHandler)
```
vgarvardt: port should be configurable

assets/apis/example_SingleDefinition.json
```diff
@@ -24,6 +24,13 @@
             }
         }
     ],
+    "circuit_breaker":{
```
vgarvardt, line 27: Shall we make it a plugin, instead of changing proxy definition?

cmd/janus/init.go
```diff
@@ -58,6 +63,26 @@ func initDistributedTracing() io.Closer {
 	return closer
 }
 
+func initCircuitBreaker() {
+	// Setup metrics for circuit breaker
+	c, err := plugins.InitializeStatsdCollector(&plugins.StatsdCollectorConfig{
+		StatsdAddr: globalConfig.Stats.DSN,
+		Prefix:     globalConfig.Stats.Prefix,
```
vgarvardt: Current `stats-go` version used in Janus is 0.4.3, but as of 0.5.0 prefix became part of the DSN. So maybe it makes sense to update stats library in this PR. Moreover - even in 0.4.3 DSN can be different from `statsd` - see https://github.com/hellofresh/stats-go#key-features

docs/misc/circuit_breaker.md
```diff
@@ -0,0 +1,41 @@
+### Circuit breaker
+
+Circuit breaker are enabled by default for each endpoint and the settings can be adjusted globally, but also per API definition.
+
+#### Configuration
+Set the following environment variables to overwrite the settings globally:
```
vgarvardt, line 6: IMO we must not allow overriding CB settings globally - this should be proxy-related setting

pkg/loader/api_loader_test.go
```diff
@@ -34,6 +33,28 @@ var tests = []struct {
 			"Content-Type": "application/json; charset=utf-8",
 		},
 		expectedCode: http.StatusOK,
+		cbFunc: func(t *testing.T) ConfigureCircuitBreakerFunc {
+			return func(name string, c CircuitBreakerConfig) {
+				if name == "" {
+					t.Fatalf("unexpected circuit breaker timeout: %d", c.Timeout)
```
vgarvardt: do non mix `github.com/stretchr/testify` asserts with "native" ones

assets/apis/example_SingleDefinition.json
italolelis: it needs to be part of the definition because we use it on the transport layer :/

cmd/janus/init.go
italolelis: :+1:

cmd/janus/init.go
italolelis: Yes, it should be configurable I. The circuit breaker config section

docs/misc/circuit_breaker.md
italolelis: What if I want to configure one CB for all proxies and I have more than 300 proxies? I’ll have to go one by one and configure be same thing on all?

pkg/loader/api_loader_test.go
italolelis: Yes please, let’s use assert everywhere

assets/apis/example_SingleDefinition.json
```diff
@@ -24,6 +24,13 @@
...
+      "timeout": 1,
```
rafaeljesus: Let's put more realistic values here 😉

assets/apis/example_SingleDefinition.json
rafaeljesus: we started as a plugin, but realised it really should be part of the definition

rafaeljesus:
We need to rethink this implementation, the failing target url should be remove from the balancer when CB trips and added back when it closes, otherwise we might have healthy target services but the failed one can trip the CB. Consul manages this out of the box, but if you don't operate janus in such way the CB won't be accurate

italolelis:
@rafaeljesus can you take this task with @c4pone? It would be nice to have this feature as soon as possible, with the scenario that you just mentioned

pkg/config/specification.go
```diff
@@ -139,6 +152,13 @@ func init() {
 	viper.SetDefault("tracing.jaeger.serviceName", serviceName)
 	viper.SetDefault("tracing.jaeger.bufferFlushInterval", "1s")
 	viper.SetDefault("tracing.jaeger.logSpans", false)
+	viper.SetDefault("circuitBreaker.timeout", hystrix.DefaultTimeout)
+	viper.SetDefault("circuitBreaker.MaxConcurrent", hystrix.DefaultMaxConcurrent)
```
rafaeljesus, line 156: it should be `maxConcurrent`

pkg/proxy/transport.go
```diff
@@ -116,9 +126,21 @@ func NewOutChain(out ...OutLink) OutChain {
...
-	if err != nil {
+	err = hystrix.Do(s.breaker.ID, func() error {
+		resp, err = http.DefaultTransport.RoundTrip(req)
+		if err != nil {
+			return err
+		}
+
+		// treat 500 and above as errors for the sake of the circuit breaker
+		if resp.StatusCode >= http.StatusInternalServerError {
+			return errors.New("internal server error")
```
rafaeljesus, line 137: `fmt.Errorf("http response error: %v", res.StatusCode)`

pkg/proxy/transport.go
```diff
@@ -61,11 +70,12 @@ type InChain []router.Constructor
...
 	statsClient client.Client
 	outbound    OutChain
+	breaker     Breaker
 }
 
 // NewTransport creates a new instance of Transport
-func NewTransport(statsClient client.Client, outbound OutChain) *Transport {
-	return &Transport{statsClient, outbound}
+func NewTransport(statsClient client.Client, outbound OutChain, breaker Breaker) *Transport {
+	return &Transport{statsClient: statsClient, outbound: outbound, breaker: breaker}
```
rafaeljesus, line 78: given all of them has the same name:
```go
&Transport{statsClient, outbound, breaker}
```

c4pone:
@italolelis I will close this pr for now and will think about a better solution next week together with @rafaeljesus. We will have a look into how we can trip the circuit breaker for each individual target without tripping for the whole endpoint. Also we were thinking about using a different circuit breaker library as hystrix-go is not the way to go.

PR #270, opened 2018-03-01
Add token metrics

italolelis:
## What does this PR do?

Adds more metrics to token validation. If it's a JWT we can be very precise about which kind of error to track on the metrics. This will allow you to build dashboards with metrics categories for each type of error.

### DIFF
--- a/pkg/metrics/metrics_context.go
+++ b/pkg/metrics/metrics_context.go
@@ -1,0 +1,26 @@
+package metrics
+
+import (
+	"context"
+
+	stats "github.com/hellofresh/stats-go"
+)
+
+type statsKeyType int
+
+const statsKey statsKeyType = iota
+
+// NewContext returns a context that has a stats Client
+func NewContext(ctx context.Context, client stats.Client) context.Context {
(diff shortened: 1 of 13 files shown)


pkg/metrics/metrics_context.go
```diff
@@ -0,0 +1,26 @@
...
+// NewContext returns a context that has a stats Client
+func NewContext(ctx context.Context, client stats.Client) context.Context {
+	return context.WithValue(ctx, statsKey, client)
+}
+
+// WithContext returns a stats Client with as much context as possible
+func WithContext(ctx context.Context) stats.Client {
+	ctxStats, ok := ctx.Value(statsKey).(stats.Client)
+	if !ok {
+		ctxStats, _ := stats.NewClient("noop://", "")
```
vgarvardt, line 22: Can you bump `stats-go` to the latest version here or should we do this in separate PR?

italolelis: Lets do it in a separate PR

vgarvardt (approved):

PR #281, opened 2018-03-10
Added leeway support for JWT date fields

vgarvardt:
Leeway helps to solve problems with clock skew on different instances. This PR introduces quick and dirty solution that is loosely based on https://github.com/dgrijalva/jwt-go/issues/131 and that should be replaced with the `github.com/dgrijalva/jwt-go` solution as soon as 4.0 with leeway support will be released https://github.com/dgrijalva/jwt-go/pull/248.

### DIFF
--- a/pkg/plugin/oauth2/oauth.go
+++ b/pkg/plugin/oauth2/oauth.go
@@ -78,7 +78,7 @@ type IntrospectionSettings struct {
 type TokenStrategy struct {
 	Name     string      `bson:"name" json:"name"`
 	Settings interface{} `bson:"settings" json:"settings"`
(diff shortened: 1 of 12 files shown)


pkg/plugin/oauth2/oauth.go
```diff
@@ -78,6 +78,10 @@ type IntrospectionSettings struct {
 type TokenStrategy struct {
 	Name     string      `bson:"name" json:"name"`
 	Settings interface{} `bson:"settings" json:"settings"`
+	// TODO: this should become part of the settings, but fot "jwt" strategy we expect array of signing methods
```
italolelis: `for` instead of `fot`

rafaeljesus (approved):

PR #295, opened 2018-03-27
Fix bug for introspection in oauth2 plugin

coolerfall:
## What does this PR do?
Fix bug for introspection in oauth2 plugin:
1. http.Request.Form should be `make` before using.
2. The tag `mapstructure` should be set in IntrospectionSettings.

### DIFF
--- a/pkg/plugin/oauth2/oauth.go
+++ b/pkg/plugin/oauth2/oauth.go
@@ -66,12 +66,12 @@ type corsMeta struct {
 
 // IntrospectionSettings represents the settings for introspection
 type IntrospectionSettings struct {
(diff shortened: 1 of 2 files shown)


italolelis (approved):
@Coolerfall Thanks for your contribution. Can you just tell me why do we need the mapstructure tag in the `IntrospectionSettings` struct?

coolerfall:
@ italolelis I checked [API.feature](https://github.com/hellofresh/janus/blob/master/features/API.feature), and I saw that you guys use lower case config field, such as `use_auth_header`, `auth_header_type`. However, it did not work when I configured `introspection` like this. I checked the source code to find that the setting field is not tagged with `mapstructure`, so I add the tag, and it works.

PR #305, opened 2018-04-12
Mongo pooling

italolelis:
## What does this PR do?

- Lots of refactoring on how the server is created and handled (Preparing to have multiple providers)
- Added watch to mongodb, filesystem and in memory providers


pkg/api/mongodb_repository.go
```diff
@@ -12,12 +16,56 @@ const (
...
+		for {
+			select {
+			case <-refreshTicker.C:
+				defs, err := r.FindAll()
+				if err != nil {
+					log.WithError(err).Error("Failed to get configurations on watch")
+					continue
+				}
+
+				cfgChan <- ConfigurationChanged{
```
vgarvardt, line 92: Are you calculating diff between current defs list and this one on the router/reciever level or just reload the router with every timer tick?

**UPD** I think I found the answer https://github.com/hellofresh/janus/pull/305/files#diff-09856fe9becddf0199651f451409356aR173

pkg/proxy/register.go
```diff
@@ -28,6 +28,11 @@ func NewRegister(router router.Router, params Params) *Register {
 	return &Register{router, params}
 }
 
+// UpdateRouter updates the reference to the router. This is useful to reload the mutex
```
vgarvardt: `reload the mutex` - what is `mutex` in this case? maybe `mux`?

pkg/api/mongodb_repository.go
italolelis: Yes, I calculate the diff with `reflect.DeepEquals` like `traefik` does

cmd/janus/context.go
```diff
@@ -0,0 +1,22 @@
...
+
+// ContextWithSignal create a context cancelled when SIGINT or SIGTERM are notified
+func ContextWithSignal(ctx context.Context) context.Context {
+	newCtx, cancel := context.WithCancel(ctx)
+	signals := make(chan os.Signal)
+	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		select {
+		case <-signals:
+			cancel()
```
rafaeljesus, line 18: you can also `close(signals)`

pkg/api/mongodb_repository.go
```diff
@@ -12,12 +16,56 @@ const (
...
+	log.Debug("Connected to MongoDB")
+	session.SetMode(mgo.Monotonic, true)
+
+	return &MongoRepository{Session: session, refreshTime: refreshTime}, nil
+}
+
+// Close terminates the session.  It's a runtime error to use a session
+// after it has been closed.
+func (r *MongoRepository) Close() error {
+	r.Session.Close()
```
rafaeljesus, line 42: shouldn't we `return r.Session.Close()`?

pkg/api/repository.go
```diff
@@ -17,13 +18,15 @@ const (
 
 // Repository defines the behavior of a proxy specs repository
 type Repository interface {
+	io.Closer
```
rafaeljesus, line 21: 👍

pkg/api/repository.go
```diff
@@ -17,13 +18,15 @@ const (
...
 	FindAll() ([]*Definition, error)
 	FindByName(name string) (*Definition, error)
 	FindByListenPath(path string) (*Definition, error)
 	Exists(def *Definition) (bool, error)
 	Add(app *Definition) error
 	Remove(name string) error
 	FindValidAPIHealthChecks() ([]*Definition, error)
+	Watch(ctx context.Context, cfgChan chan<- ConfigurationChanged)
```
rafaeljesus: we could be a interface as well `Watcher`

rafaeljesus: which exposes `Watch` method

pkg/server/server.go
```diff
@@ -0,0 +1,259 @@
...
+
+// Server is the Janus server
+type Server struct {
+	server                *http.Server
+	provider              api.Repository
+	register              *proxy.Register
+	defLoader             *loader.APILoader
+	currentConfigurations []*api.Definition
+	configurationChan     chan api.ConfigurationChanged
+	stopChan              chan bool
```
rafaeljesus: ideally it should be a struct

pkg/server/server.go
```diff
@@ -0,0 +1,259 @@
...
+	go func(ctx context.Context) {
+		<-ctx.Done()
+		if ctx.Err() == context.Canceled {
+			return
+		} else if ctx.Err() == context.DeadlineExceeded {
+			panic("Timeout while stopping janus, killing instance ✝")
+		}
+	}(ctx)
+
+	close(s.stopChan)
```
rafaeljesus: Usually closing resources stays in the top wrapped by defer e.g:

```go
func (s *Server) Close() error {
  defer close(s.stopChan)
  defer close(s.configurationChan)

  // code goes here
}
```

pkg/server/server.go
```diff
@@ -0,0 +1,259 @@
...
+		log.Info("Stopping server gracefully")
+		s.Close()
+	}()
+	go func() {
+		if err := s.startHTTPServers(); err != nil {
+			log.WithError(err).Fatal("Could not start http servers")
+		}
+	}()
+	go func() {
+		s.listenProviders(s.stopChan)
```
rafaeljesus: it could be just `go s.listenProviders(s.stopChan)`

pkg/server/server.go
```diff
@@ -0,0 +1,259 @@
...
+		defer s.Close()
+		<-ctx.Done()
+		log.Info("I have to go...")
+		reqAcceptGraceTimeOut := time.Duration(s.globalConfig.GraceTimeOut)
+		if reqAcceptGraceTimeOut > 0 {
+			log.Infof("Waiting %s for incoming requests to cease", reqAcceptGraceTimeOut)
+			time.Sleep(reqAcceptGraceTimeOut)
+		}
+		log.Info("Stopping server gracefully")
+		s.Close()
```
rafaeljesus: extra call of Close method

pkg/api/mongodb_repository.go
italolelis: `r.Session.Close()` from mongo returns void, so I can't return it

rafaeljesus (approved):

PR #309, opened 2018-04-19
Plugin Circuit Breaker

### LINKED ISSUE #301: Implement Circuit Breaker for Targets
The circuit breaker implementation should be by targets of each proxy.

- The CB configuration should be by proxy
- The CB should be implemented on the transport layer
- When the CB trips we should notify the metrics system (statsd, prometheus, etc)
- When the CB returns probably the retry #300 will be called
- The CB that we should use is probably [hystrix-go](https://github.com/afex/hystrix-go)

italolelis:
## What does this PR do?
- Adds the CB plugin
- It is based on [hystrix](https://github.com/Netflix/Hystrix) from Netflix
- It uses [hystrix-go](https://github.com/afex/hystrix-go)
- It contains examples with a hystrix-dashboard
- Changes the plugin signature to pass the API definition

Closes #301


docs/plugins/cb.md
```diff
@@ -0,0 +1,32 @@
...
+        "predicate": "statusCode == 0 || statusCode >= 500"
+    }
+}
+```
+
+| Configuration | Description |
+| timeout                     | Timeout that the CB will wait till the request responds |
+| max_concurrent_requests     | How many commands of the same type can run at the same time |
+| error_percent_threshold     | Causes circuits to open once the rolling measure of errors exceeds this percent of requests |
+| error_percent_threshold     | Is the minimum number of requests needed before a circuit can be tripped due to health |
```
vgarvardt: `error_percent_threshold` second time

pkg/plugin/cb/middleware.go
```diff
@@ -0,0 +1,66 @@
...
+
+			err = hystrix.Do(def.Name, func() error {
+				m := httpsnoop.CaptureMetrics(handler, w, r)
+				params := make(map[string]interface{}, 8)
+				params["statusCode"] = m.Code
+				params["request"] = r
+
+				result, err := expression.Evaluate(params)
+				if err != nil {
+					return errors.New("cannot evaluate the expression")
```
vgarvardt, line 50: can not

pkg/plugin/cb/setup.go
```diff
@@ -0,0 +1,102 @@
...
+
+	e, ok := event.(plugin.OnStartup)
+	if !ok {
+		return errors.New("Could not convert event to startup type")
+	}
+
+	if e.Config.Stats.DSN != "" {
+		logger.WithField("metrics_dsn", e.Config.Stats.DSN).Debug("Statsd metrics enabled")
+		c, err := plugins.InitializeStatsdCollector(&plugins.StatsdCollectorConfig{
+			StatsdAddr: e.Config.Stats.DSN,
```
vgarvardt: if this is DSN from global stats then there is no guarantee that it is `statsd` - can be `noop`, `log`, `memory`, `prometeus` (soon)

pkg/plugin/cb/setup.go
```diff
@@ -0,0 +1,102 @@
...
+			Prefix:     statsdPrefix,
```
vgarvardt: In statsd prefix is the global thing, that means all metrics will be published to the root of all metrics with the set prefix, `hystrix` in this case. Out stats lib allows to set prefix in DSN, that allows you to set prefix depending on the environment, e.g. `api-gateway.staging`, `api-gateway.live`, etc.

pkg/plugin/cb/setup.go
rafaeljesus: so we can use `url.Parse`:

```go
u, _ := url.Parse(e.Config.Stats.DSN)
switch u.Scheme {
  case "statsd":
  case "prometeus":
}
```

pkg/plugin/cb/setup.go
rafaeljesus: hm good catch, so maybe I would go for writing our own collector given we might need to do the same for prometheus. And yes another PR

vgarvardt: Maybe we can assert our stats client type and if it is statsd - set this guy. And IMO we can expose prefix from our client implementation - then you'll need only to join `hystrix` with the current prefix, prefix may end with a dot or not - both valid.

examples/plugin-cb/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+Start making a lot of requests to `/example` and see what's going to happen on the dashboard.
+
+For all the options on how to configure this plugin please visit the [documentation](https://hellofresh.gitbooks.io/janus/plugins/cb.html) page.
+
+## Using vegeta for failure simulation
+
+You can also use [vegeta](https://github.com/tsenart/vegeta) which is a HTTP load testing tool.
+
+Here we are going to send 10 req/s during 30s. IN this time you can easily stop `service1` and see how the circuit will react.
```
rafaeljesus: `In this time ...`

examples/plugin-cb/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+```sh
+curl localhost:8080/example
+```
+
+If you take a look on the dashboard you should see something like this:
+
+![](img/hystrix-dashboard.png)
+
+## Simulating failure
```
rafaeljesus, line 25: 👍

pkg/config/specification.go
```diff
@@ -1,11 +1,11 @@
 package config
 
 import (
-	"os/user"
 	"time"
 
 	"github.com/hellofresh/logging-go"
 	"github.com/kelseyhightower/envconfig"
+	homedir "github.com/mitchellh/go-homedir"
```
rafaeljesus: we can omit `homedir` from the import

pkg/plugin/cb/middleware.go
```diff
@@ -0,0 +1,66 @@
...
+			}
+
+			expression, err := govaluate.NewEvaluableExpression(cfg.Predicate)
+			if err != nil {
+				log.WithError(err).Error("could not create an expression with this predicate")
+				handler.ServeHTTP(w, r)
+				return
+			}
+
+			err = hystrix.Do(def.Name, func() error {
```
rafaeljesus, line 42: quick question, haven't we decided to make use a map in order to avoid the issues we've had with the old implementation?

rafaeljesus: I mean have a map of the proxy name + the backend target:

- `proxy1-name:backend-service000.company.io`
- `proxy1-name:backend-service001.company.io`

Allowing us to have one circuit breaker for each target, although I don't remember when we discussed about it would be possible to do it in the middleware layer

rafaeljesus (approved):

PR #311, opened 2018-04-24
Fix markup so it will looks like a table

ivanzusko:
## What does this PR do?

This **PR** fixes the markup, so instead of this:

image

it will look like this:

image


julyate:
+1

vgarvardt (approved):

rafaeljesus (approved):

PR #313, opened 2018-04-27
Add postgres repository

whisper-bye:
add postgres repository, for example

```
create table apis
(
	json jsonb
)
;

create table oauth
(
	json jsonb
)
;
```

```
[database]
  dsn = "postgres://postgres@localhost:5432/janus"
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -214,6 +214,15 @@
   version = "v1.3.0"
 
 [[projects]]
+  branch = "master"
+  name = "github.com/lib/pq"
+  packages = [
+    ".",
+    "oid"
+  ]
+  revision = "d34b9ff171c21ad295489235aec8b6626023cd04"
+
+[[projects]]
   name = "github.com/magiconair/properties"
   packages = ["."]
   revision = "d419a98cdbed11a922bf76f257b7c4be79b50e73"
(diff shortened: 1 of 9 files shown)


Gopkg.lock
```diff
@@ -213,6 +213,15 @@
   revision = "f611eb38b3875cc3bd991ca91c51d06446afa14c"
   version = "v1.3.0"
 
+[[projects]]
+  branch = "master"
+  name = "github.com/lib/pq"
```
vgarvardt, line 218: I'd rather used https://github.com/jackc/pgx instead of `lib/pq`

pkg/api/postgres_repository.go
```diff
@@ -0,0 +1,149 @@
+package api
+
+import (
+	"context"
+	"encoding/json"
+	log "github.com/sirupsen/logrus"
+	"time"
```
vgarvardt: format inports pls

vgarvardt (commented):
For relational DB it makes sense to use DB features, e.g. extract proxy/oauth name to a separate column with primary key to let the DB validate name uniqueness and make update/delete queries simplier.

Gopkg.lock
rafaeljesus: @whisper-bye thanks for your contribution, `upper/db` is too much for our use case, we need a tinier one as. In addition to that ideally if postgres being supported we don't need the polling implementation given PG has notify/listen mechanism, so the PG implementation would have this feature

whisper-bye: thanks, close this PR

PR #318, opened 2018-05-07
Improve cmds

italolelis:
## What does this PR do?
Uses the cobra cmd as it should be. This means that now we have a new command called `janus start` which starts the web server


Dockerfile
```diff
@@ -27,4 +27,4 @@ RUN apk add --update curl && \
 HEALTHCHECK --interval=5s --timeout=5s --retries=3 CMD curl -f http://localhost:8081/status || exit 1
 
 EXPOSE 8080 8081 8443 8444
-ENTRYPOINT ["/janus"]
+ENTRYPOINT ["/janus", "start"]
```
vgarvardt, line 30: is there a way to make "start" default command and omit it here?

italolelis: Not really a good idea. When we use a CLI cmd we expect it to show the help on the root cmd, right? It's like this with `curl`, `vegeta`, `http`, that's why we have the `start` command

rafaeljesus (approved):

Issue #320, opened 2018-05-10
Is there a way to increase request ttl for api definition ? Got http: proxy error: context canceled


jzyinq:
Is there a way to increase request ttl ? 

Something like `upstream_[connect|send|read]_timeout` ? 

**Reproduction Steps:**

1. Make a request through Janus that will last longer than 50 seconds.

**Expected behavior:**

Requests is handled properly in his own time.

**Observed behavior:**

Request is terminated with Janus error:

`http: proxy error: context canceled`

**Janus version:** 3.7.0-rc.14 Docker image

jzyinq:
@italolelis could you refer to this ?

italolelis:
Hi @jzyinq, 

Thanks for reporting this. At the moment we don't have this configuration per proxy. But we have a task to add a plugin that you can configure the timeout for each endpoint. We should have this functionality soon.

jzyinq:
Thanks for reply @italolelis,

Sorry for bothering but how soon is soon ? ;) 

Could you guide me then how to set this globally for now ?

rafaeljesus:
@jzyinq in fact we are going to add this plugin to actually decrease the current time from 60s to 30s at most, in very exceptional cases a request should take that much, if the downstream service does not respond quickly the proxy should just kill the connection. What's your use case?

jzyinq:
@rafaeljesus We got some requests that streams very big payloads of data and that takes about 60-120 seconds at the moment. So in our case response is quick but reading it takes more time and it looks like Janus is waiting for whole payload before letting it outside.

Do you have any thoughts how we can tackle this problem ?

italolelis:
@jzyinq I see. Janus should have the capabilities for you to specify a timeout per endpoint. I'll try to implement this asap.

quba:
@italolelis is there any eta? We have a pending feature that is blocked because of this issue and before the next product release we have to decide whether we go with Janus or Kong. That's the only blocker for us currently. Thanks! :)

quba:
Wow, that's fast. 👍

italolelis:
@quba The new version with the timeouts is out, please check if it attends your needs

quba:
Great! We should be able to provide some feedback within next 2 weeks.

glothriel:
Hi @italolelis 
[...]

PR #323, opened 2018-05-28
Add Timeouts functionality

italolelis:
## What does this PR do?

- Improves the Makefile
- Removes the `-race` flag from tests as this breaks on `alpine`, [this](https://github.com/golang/go/issues/14481) is golang issue
- Removes the `enable_load_balancing` property which was not used
- Removes the `ReadOnly` configuration for the admin API
- Removes the `Debug` configuration from the configuration as this was not used
- Makes the transport layer less smart and leave the stats tracking to the stats middleware
- Moves the balancer logic to its own package
- Adds an alias for the `roundrobin` load balancing which is called `rr` 
- Adds request and server timeouts and solves #320. Now we'll be able to configure timeouts per endpoits as well as timeouts for Janus server. Everything comes with defaults so this should not affect productions servers at HelloFresh. The new configurations are:

For Janus server:

```toml
#[respondingTimeouts]
# readTimeout is the maximum duration for reading the entire request, including the body.
#
# Optional
# Default: "0s"
#
# readTimeout = "5s"

# writeTimeout is the maximum duration before timing out writes of the response.
#
# Optional
# Default: "0s"
#
# writeTimeout = "5s"

# idleTimeout is the maximum duration an idle (keep-alive) connection will remain idle before closing itself.
#
# Optional
# Default: "180s"
#
# idleTimeout = "360s"
```

Per endpoint:

```json
{
"forwarding_timeouts": {"dial_timeout": "30s", "response_header_timeout": "60s"} 
}
```


pkg/proxy/balancer/rr.go
```diff
@@ -0,0 +1,35 @@
...
+		mu      sync.RWMutex
+	}
+)
+
+// NewRoundrobinBalancer creates a new instance of Roundrobin
+func NewRoundrobinBalancer() *RoundrobinBalancer {
+	return &RoundrobinBalancer{}
+}
+
+// Elect backend using roundrobin strategy
```
vgarvardt: I fixed this comment in the PR that was just merged =)

pkg/proxy/balancer/weight.go
```diff
@@ -0,0 +1,50 @@
...
+	// ErrZeroWeight is used when there a zero value weight was given
+	ErrZeroWeight = errors.New("invalid backend, weight 0 given")
+)
+
+// NewWeightBalancer creates a new instance of Roundrobin
+func NewWeightBalancer() *WeightBalancer {
+	return &WeightBalancer{}
+}
+
+// Elect backend using roundrobin strategy
```
vgarvardt: I fixed this comment in the PR that was just merged - `roundrobin` -> `weight` =)

pkg/proxy/balancer/weight.go
```diff
@@ -0,0 +1,50 @@
...
+	WeightBalancer struct{}
+)
+
+var (
+	// ErrZeroWeight is used when there a zero value weight was given
+	ErrZeroWeight = errors.New("invalid backend, weight 0 given")
+)
+
+// NewWeightBalancer creates a new instance of Roundrobin
+func NewWeightBalancer() *WeightBalancer {
```
vgarvardt: `Roundrobin` -> `WeightBalancer`

pkg/proxy/balancer/weight.go
```diff
@@ -0,0 +1,50 @@
...
+	totalWeight := 0
+	for _, host := range hosts {
+		totalWeight += host.Weight
+	}
+
+	if totalWeight <= 0 {
+		return nil, ErrZeroWeight
+	}
+
+	r := rand.Intn(totalWeight)
```
vgarvardt, line 38: I think we should add something like this to init rand seed

```
func init() {
	rand.Seed(time.Now().UnixNano())
} 
```

pkg/proxy/balancer/weight.go
```diff
@@ -0,0 +1,50 @@
...
+	// WeightBalancer balancer
+	WeightBalancer struct{}
+)
+
+var (
+	// ErrZeroWeight is used when there a zero value weight was given
+	ErrZeroWeight = errors.New("invalid backend, weight 0 given")
+)
+
+// NewWeightBalancer creates a new instance of Roundrobin
```
vgarvardt: `Roundrobin` -> `WeightBalancer`

vgarvardt:
@italolelis can you pls split this PR into 2 or 3 to make changes more granular and review a bit easier? 

E.g. extract `-race`, `enable_load_balancing` and `Debug` related changes into the first PR, balancing changes/refactoring into another and Timeout functionality into the third one.

pkg/proxy/balancer/weight.go
italolelis: It's being initialized on the balancer.go file. It's for the whole package

italolelis:
@vgarvardt The timeouts and balancer transport refactors have to be on the same PR, they are dependent. I could split the race stuff but they are tiny, that's why they are in the same PR. If you want I can split the `-race`, `enable_load_balancing` and `Debug` but not the other two

docs/config/proxy.md
```diff
@@ -4,8 +4,9 @@
...
 | preserve_hosts        | Enable the [preserve host](/docs/proxy/preserve_host_property.md) definition           |
 | listen_path           | Defines the [endpoint](/docs/proxy/request_uri.md) that will be exposed in Janus       |
-| upstreams             | Defines the OAuth provider's [endpoints](/docs/proxy/upstreams.md) for the request   |
+| upstreams             | Defines the [endpoints](/docs/proxy/upstreams.md) that the request will be forwarded to|
 | strip_path            | Enable the [strip URI](/docs/proxy/strip_uri_property.md) rule on this proxy           |
-| enable_load_balancing | Enable load balancing for this proxy                                                   |
 | methods               | Defines which [methods](/docs/proxy/request_http_method.md) are enabled for this proxy |
 | hosts                 | Defines which [hosts](/docs/proxy/request_http_header.md) are enabled for this proxy   |
+| forwarding_timeouts.dial_timeout | The amount of time to wait until a connection to a backend server can be established. Defaults to 30 seconds. If zero, no timeout exists |
```
vgarvardt: Can you point here that interval is in parsable `time.Duration` add link to Parse function with examples in go docs? Now the format is not clear here and in the parameter bellow.

pkg/config/specification.go
```diff
@@ -13,7 +13,6 @@ import (
 // Specification for basic configurations
 type Specification struct {
 	Port                 int           `envconfig:"PORT"`
-	Debug                bool          `envconfig:"DEBUG"`
```
vgarvardt, line 16: 👍

pkg/config/specification.go
```diff
@@ -132,6 +138,8 @@ func init() {
 	viper.SetDefault("backendFlushInterval", "20ms")
 	viper.SetDefault("requestID", true)
 
+	viper.SetDefault("respondingTimeouts.IdleTimeout", 180*time.Second)
```
vgarvardt, line 141: should not it be `respondingTimeouts.idleTimeout`? `idle` not capital - test is the best way to check it =)

italolelis: Both work, but I'll add a test for it anyways

vgarvardt (approved):
ltgm

docs/config/proxy.md
```diff
@@ -4,8 +4,9 @@
...
 | listen_path           | Defines the [endpoint](/docs/proxy/request_uri.md) that will be exposed in Janus       |
-| upstreams             | Defines the OAuth provider's [endpoints](/docs/proxy/upstreams.md) for the request   |
+| upstreams             | Defines the [endpoints](/docs/proxy/upstreams.md) that the request will be forwarded to|
 | strip_path            | Enable the [strip URI](/docs/proxy/strip_uri_property.md) rule on this proxy           |
-| enable_load_balancing | Enable load balancing for this proxy                                                   |
 | methods               | Defines which [methods](/docs/proxy/request_http_method.md) are enabled for this proxy |
 | hosts                 | Defines which [hosts](/docs/proxy/request_http_header.md) are enabled for this proxy   |
+| forwarding_timeouts.dial_timeout | The amount of time to wait until a connection to a backend server can be established. Defaults to 30 seconds. If zero, no timeout exists. You must use any format that is compatible with [time.Duration](https://golang.org/pkg/time/#Duration) |
```
rafaeljesus, line 11: We might have side effects in some HF proxies I guess, can we make it 60s like it was before? 😢

pkg/config/specification.go
```diff
@@ -26,17 +25,24 @@ type Specification struct {
...
 	Cluster              Cluster
+	RespondingTimeouts   RespondingTimeouts
 }
 
 // Cluster represents the cluster configuration
 type Cluster struct {
 	UpdateFrequency time.Duration `envconfig:"BACKEND_UPDATE_FREQUENCY"`
 }
 
+// RespondingTimeouts contains timeout configurations for incoming requests to the Traefik instance.
```
rafaeljesus: `Traefik` typo

docs/config/proxy.md
vgarvardt: I think we should provide a "migration" script in upgrade docs to fill proxies with old default values.

rafaeljesus: > Dial was always 30s

ah that's fine then

rafaeljesus (approved):

PR #325, opened 2018-05-31
Update check.go

remisb:
## What does this PR do?
Minor CLI output formatting fixed for "janus check" command. Added line brake at the end.

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -24,6 +24,6 @@ func RunCheck(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	cmd.Printf("The configuration file is valid")
+	cmd.Printf("The configuration file is valid\n")
 	return nil
 }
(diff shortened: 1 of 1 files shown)


cmd/check.go
```diff
@@ -24,6 +24,6 @@ func RunCheck(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	cmd.Printf("The configuration file is valid")
+	cmd.Printf("The configuration file is valid\n")
```
vgarvardt: `cmd.Println(...)` would be a bit better in this case IMO

vgarvardt (commented):
Thank you for your contribution. I left a small comment. Travis is very unstable in the last several days - I'll restart failing tests as soon as you commit a fix.

vgarvardt (approved):

PR #328, opened 2018-06-14
PT-1163 Fixed memory leak caused by creating eternal goroutines for every new transport

vgarvardt:
## What does this PR do?

This PR solves the problem with memory leaking on each reouter reload. The problem was in the transport - each definition created own transport, and each launched infinite go-routine. The solution is the following: do not create transport for each route, reuse one from the registry instead (registry key is being built from transport parameters), and do not launch go-routine, let transport retire "keep-alive" connections using internal transport functionality.

In addition to this there are couple minor changes I did during issue investigation:
- added `pprof` http handlers to private port, can be enabled only with command arguments
- replaced unmaintained `gopkg.in/mgo.v2` with community-maintained fork `github.com/globalsign/mgo` - profiling shown that mongo stuff (like BSON-decoding) consumes quite a lot of resources for ~100 proxies loaded every second from Mongo

One known issue is still not fixed - the way I discovered this memory leak issue is that `reflect.DeepEqual()` does not work properly when comparing current loaded from Mongo data in our environment. I'm in the process of investigation, but let it be in the next PR.


pkg/proxy/transport/transport.go
```diff
@@ -53,8 +68,15 @@ func New(opts ...Option) *http.Transport {
...
-	if t.closeIdleConnsPeriod == 0 {
-		t.closeIdleConnsPeriod = DefaultCloseIdleConnsPeriod
+	if t.idleConnTimeout == 0 {
+		t.idleConnTimeout = DefaultIdleConnTimeout
+	}
+
+	// let's try to get the cached transport from registry, since there is no need to create lots of
+	// transports with the same configuration
+	hash := t.hash()
+	if tr, ok := registry[hash]; ok {
```
italolelis: I think this might cause race conditions, right? If we create transactions inside goroutines, this can break. Did you run it with the `--race` flag to check?

vgarvardt: Good catch! No, I did not check it, will do and add thread-safe for registry.

pkg/proxy/transport/registry.go
```diff
@@ -0,0 +1,34 @@
...
+	r.store = make(map[string]*http.Transport)
+
+	return r
+}
+
+func (r *registry) get(key string) (*http.Transport, bool) {
+	r.RLock()
+	defer r.RUnlock()
+
+	// return r.store[key] does not work here, says too few argument to return
```
rafaeljesus, line 24: 😮

rafaeljesus (approved):

PR #329, opened 2018-06-19
PT-1189 Fixed routers list comparison

vgarvardt:
## What does this PR do?
This PR fixes routes list comparison. The problem was in private `proxy.Definition.middleware` field - because of this I extracted the field into new `proxy.RouterDefinition` that is now used in all plugins, routers, etc.


pkg/loader/api_loader.go
```diff
@@ -52,7 +54,7 @@ func (m *APILoader) RegisterAPI(def *api.Definition) {
 					continue
 				}
 
-				err = setup(def, plg.Config)
+				err = setup(routerDefinition, plg.Config)
```
italolelis, line 57: why are we encapsulating only the `proxy.Definition` and not the `api.Definition`?

pkg/plugin/cb/middleware.go
```diff
@@ -39,7 +38,7 @@ func NewCBMiddleware(cfg Config, def *api.Definition) func(http.Handler) http.Ha
 				return
 			}
 
-			err = hystrix.Do(def.Name, func() error {
+			err = hystrix.Do(cfg.Name, func() error {
```
italolelis, line 41: Why can't we use the proxy name?

pkg/loader/api_loader.go
vgarvardt: To fix the issue I had to remove `middleware` property from `proxy.Definition` and plugin's setup adds middlewares to a proxy. First I tried to move middleware list to `api.Definition`, but this involves too many changes - need to introduce `api.RouterDefinition` (because `api.Definition` is involved into comparison) and this couses huge code changes.

After examining the code I found that the only plugin that uses something from `api.Definition` is `cb` and it uses only definition name, so changing plugin setup parameter from `api.Definition` to `proxy.RouterDefinition` was quite easy and involved only one change - this is actually answers your second comment about introducing `name` to `cb` plugin configuration.

There is actually another solution I was thinking about - replace `reflect.DeepEqual()` with custom implementation, implement `EqualsTo()` for `proxy.Definition` and do the comparison manually, skipping dynamic fields (`middleware`), but decided not to do this - we already have problems with performance - CPU has ~10% load with refresh set to `1s` for the idle instance. Probably performance investigation anf optimisation will be the next priority.

rafaeljesus (approved):

PR #330, opened 2018-06-22
Add forwarding timeout duration marshalling

s4nji:
## What does this PR do?
Adds JSON Marshalling for ForwardingTimeout duration.

#### Ticket
https://hellofresh.atlassian.net/browse/PT-1272

### DIFF
--- a/pkg/proxy/definition_test.go
+++ b/pkg/proxy/definition_test.go
@@ -138,8 +138,8 @@ func testMarshalForwardingTimeoutsToJSON(t *testing.T) {
 	}
 	jsonDefinition, err := json.Marshal(&definition)
 	require.NoError(t, err)
-	assert.Contains(t, string(jsonDefinition), "30s")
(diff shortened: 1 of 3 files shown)


pkg/proxy/definition_test.go
```diff
@@ -118,6 +122,26 @@ func testAddMiddlewares(t *testing.T) {
...
+			},
+		},
+		ForwardingTimeouts: ForwardingTimeouts{
+			DialTimeout:           Duration(30 * time.Second),
+			ResponseHeaderTimeout: Duration(31 * time.Second),
+		},
+	}
+	jsonDefinition, err := json.Marshal(&definition)
+	require.NoError(t, err)
+	assert.Contains(t, string(jsonDefinition), "30s")
```
vgarvardt: add field name here and the line bellow as well to be alightly more specific, because we may have another duration fields in the definition, e.g. `"dial_timeout":"30s"`

vgarvardt (approved):

PR #331, opened 2018-07-02
Configurable Session Timeout

s4nji:
## What does this PR do?
Allow administration token timeout to be configurable via envconfig (`TOKEN_TIMEOUT`)
The default value remains to be 1 hour.

## Related
https://github.com/hellofresh/janus-dashboard/issues/293

### DIFF
--- a/pkg/config/specification.go
+++ b/pkg/config/specification.go
@@ -79,9 +79,10 @@ type Stats struct {
 type Credentials struct {
 	// Algorithm defines admin JWT signing algorithm.
 	// Currently the following algorithms are supported: HS256, HS384, HS512.
-	Algorithm      string `envconfig:"ALGORITHM"`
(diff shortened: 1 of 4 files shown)


pkg/config/specification.go
```diff
@@ -147,6 +148,7 @@ func init() {
 	viper.SetDefault("web.tls.port", "8444")
 	viper.SetDefault("web.tls.redirect", true)
 	viper.SetDefault("web.credentials.algorithm", "HS256")
+	viper.SetDefault("web.credentials.timeout", time.Hour)
```
vgarvardt, line 151: Pls add this parameter to https://github.com/hellofresh/janus/blob/master/janus.sample.toml as well, just to have it documented.

vgarvardt (approved):

rafaeljesus (approved):

Issue #338, opened 2018-07-17
panic: runtime error with OAuth plugin 


jzyinq:
Janus docker images fails randomly on startup while loading OAuth plugin. 

I've checked both `3.7.2-rc.1` and `3.7.1-rc.21` images.

**Reproduction Steps:**

I've got this problem on our internal configuration, but it occurs also on `front-proxy-auth` example.

Just change image in docker-compose to one of listed above -> https://github.com/hellofresh/janus/blob/master/examples/front-proxy-auth/docker-compose.yml#L5 and run example.

After Janus starts it randomly fails with error:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x90ad0f]

goroutine 1 [running]:
github.com/hellofresh/janus/pkg/proxy.(*Register).Add(0x0, 0xc4204bc860, 0x0, 0x0)
        /go/src/github.com/hellofresh/janus/pkg/proxy/register.go:59 +0x10f
github.com/hellofresh/janus/pkg/plugin/oauth2.(*OAuthLoader).registerRoutes(0xc420434110, 0xc4204ff400)
        /go/src/github.com/hellofresh/janus/pkg/plugin/oauth2/loader.go:123 +0x2fb
github.com/hellofresh/janus/pkg/plugin/oauth2.(*OAuthLoader).RegisterOAuthServers(0xc420434110, 0xc420434158, 0x1, 0x1, 0xce7460, 0xc420458f40)
        /go/src/github.com/hellofresh/janus/pkg/plugin/oauth2/loader.go:71 +0xd9
github.com/hellofresh/janus/pkg/plugin/oauth2.(*OAuthLoader).LoadDefinitions(0xc420434110, 0xce7460, 0xc420458f40)
        /go/src/github.com/hellofresh/janus/pkg/plugin/oauth2/loader.go:26 +0x7f
github.com/hellofresh/janus/pkg/plugin/oauth2.onStartup(0xbc9e00, 0xc42045d700, 0x0, 0x0)
        /go/src/github.com/hellofresh/janus/pkg/plugin/oauth2/setup.go:109 +0x517
github.com/hellofresh/janus/pkg/plugin.EmitEvent(0xc57a15, 0x7, 0xbc9e00, 0xc42045d700, 0x0, 0x0)
        /go/src/github.com/hellofresh/janus/pkg/plugin/plugin.go:95 +0x169
[...]

PR #339, opened 2018-07-19
PT-1479 Fixed server initialisation race condition

vgarvardt:
## What does this PR do?
When loading proxy definitions from filesystem proxy register may be uninitialised in some cases because of race condition. I extracted proxy register initialisation outside of go-routine to avoid the possibbility of it.

Fixes https://github.com/hellofresh/janus/issues/338

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -212,8 +212,8 @@
     "state",
     "timer"
   ]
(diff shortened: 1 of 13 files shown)


rafaeljesus:
:shipit:

rafaeljesus (approved):

PR #344, opened 2018-08-14
Fixing a build failure when all modules are built.

amukherj:
## What does this PR do?
Accessing `client.Organizations.ListUserTeams` is incorrect. The correct way to get user teams is:

    client.Teams.ListUserTeams

### DIFF
--- a/pkg/jwt/github/client.go
+++ b/pkg/jwt/github/client.go
@@ -46,7 +46,7 @@ func (c *client) Teams(httpClient *http.Client) (OrganizationTeams, error) {
 	organizationTeams := OrganizationTeams{}
 
 	for nextPage != 0 {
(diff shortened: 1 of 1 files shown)


s4nji:
Hello! Thanks for the PR, seems like it is indeed the correct way to obtain the user team in the latest version of `google/go-github`:

https://github.com/google/go-github/blob/b1f138353a6287df5156971b3963e9107eca06a1/github/orgs_teams.go#L382
https://github.com/google/go-github/blob/cc20b8df44a10cf7439649ec5a2192f06e97fe43/github/teams.go#L356

I'm closing this pull request since it requires `Gopkg.lock` to be updated as well, it will be done in https://github.com/hellofresh/janus/pull/410.

PR #367, opened 2018-10-25
Fix table layout by escaping the `||` character

zianwar:
## What does this PR do?
While I was reading the docs on [Gitbook](https://hellofresh.gitbooks.io/janus/plugins/cb.html), I noticed that the table was broken and it seems that the `||` needs to be escaped.

[https://i.imgur.com/J3jX95E.png](https://i.imgur.com/J3jX95E.png)


vgarvardt (approved):
Thank you for the fix!

PR #369, opened 2018-11-05
PT-1875 Add OpenCensus

s4nji:
This PR adds metric collection and tracing with [OpenCensus][1].

[1]: https://opencensus.io


cmd/init.go
```diff
@@ -80,3 +87,83 @@ func initStatsClient() {
...
+	case "prometheus":
+		err = initPrometheusExporter()
+		break
+	default:
+		logger.Info("Unsupported or invalid stats exporter was specified")
+		return
+	}
+
+	if err != nil {
+		logger.Warn("Failed initialising stats exporter")
```
vgarvardt: probably you can add `.WithError(err)` here, just to have all the info in the single log entry, even if you already logged it previously

cmd/init.go
```diff
@@ -80,3 +87,83 @@ func initStatsClient() {
...
+	vv := append(ochttp.DefaultServerViews, obs.AllViews...)
+
+	if err := view.Register(vv...); err != nil {
+		log.WithError(err).Warn("Failed to register server views")
+	}
+}
+
+func initPrometheusExporter() (err error) {
+	obs.PromExporter, err = prometheus.NewExporter(prometheus.Options{
+		Namespace: "janus",
```
vgarvardt: namespace can be configurable IMO, with `janus` as default value

pkg/jwt/handler.go
```diff
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	"go.opencensus.io/trace"
+
```
vgarvardt: this extra space-line is not needed here

pkg/plugin/basic/handler.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net/http"
 
+	"go.opencensus.io/trace"
+
```
vgarvardt: this extra space-line is not needed here

pkg/plugin/oauth2/handlers.go
```diff
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"net/http"
 
+	"go.opencensus.io/trace"
+
```
vgarvardt: this extra space-line is not needed here

pkg/plugin/oauth2/jwt_manager.go
```diff
@@ -3,9 +3,13 @@ package oauth2
 import (
 	"context"
 
+	"go.opencensus.io/stats"
+	"go.opencensus.io/tag"
+
```
vgarvardt: this extra space-line is not needed here

pkg/plugin/oauth2/middleware.go
```diff
@@ -5,10 +5,12 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/hellofresh/stats-go/bucket"
+	"go.opencensus.io/stats"
```
vgarvardt: this extra space-line is not needed here

pkg/proxy/reverse_proxy.go
```diff
@@ -7,8 +7,12 @@ import (
 	"net/url"
 	"strings"
 
+	"go.opencensus.io/stats"
+	"go.opencensus.io/tag"
+
```
vgarvardt: this extra space-line is not needed here

pkg/web/api_handler.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net/http"
 
+	"go.opencensus.io/trace"
+
```
vgarvardt: this extra space-line is not needed here

vgarvardt (commented):
Overall looks good! The only thing missing is documentation for the new features implemented - what is it, how to set it up, what values are supported. etc. =)

cmd/init.go
s4nji: Yes! I believe we have `serviceName` somewhere in the config which can be used here.

spetrashov (commented):
I see that we use different naming approach for spans. Sometimes it is`datastore.{repoName}.Add` and sometimes it's just `datastore.Add`. 
We used to publish traces with these names in the previous implementation, but since this PR replaces the tracing library, we could make span names consistent.

pkg/observability/observability.go
```diff
@@ -0,0 +1,79 @@
...
+	MOAuth2MissingHeader        = stats.Int64("opencensus.io/plugin/oauth2/missing_header_total", "Number of failed oauth2 authentication due to missing header", dimensionless)
+	MOAuth2MalformedHeader      = stats.Int64("opencensus.io/plugin/oauth2/malformed_header_total", "Number of failed oauth2 authentication due to malformed bearer header", dimensionless)
+	MOAuth2Authorized           = stats.Int64("opencensus.io/plugin/oauth2/authorized_request_total", "Number of successful and authorized oauth2 authentication", dimensionless)
+	MOAuth2Unauthorized         = stats.Int64("opencensus.io/plugin/oauth2/unauthorized_request_total", "Number of successful but unauthorized oauth2 authentication", dimensionless)
+)
+
+// AllViews aggregates the metrics
+var AllViews = []*view.View{
+	{
+		Name:        "opencensus.io/http/proxy/request_total_by_host",
```
spetrashov: Would it be helpful to have proxy request latency by the host (or, maybe even by the API Name)?
The total amount of requests by host show us only usage information, not the performance of the proxy.

vgarvardt: API name and HTTP method would be even better

pkg/observability/observability.go
```diff
@@ -0,0 +1,79 @@
...
+	AzureMonitor = "azure_monitor"
+	Datadog      = "datadog"
+	Jaeger       = "jaeger"
+	Prometheus   = "prometheus"
+	Stackdriver  = "stackdriver"
+	Zipkin       = "zipkin"
+)
+
+// PromExporter is the prometheus exporter containing HTTP handler for "/metrics"
+var PromExporter *prometheus.Exporter
```
vgarvardt:
```suggestion
var PrometheusExporter *prometheus.Exporter
```

cmd/init.go
```diff
@@ -80,3 +87,105 @@ func initStatsClient() {
...
+	case obs.Jaeger:
+		err = initJaegerExporter()
+		break
+	default:
+		logger.Info("Invalid or no tracing exporter was specified")
+		return
+	}
+
+	if err != nil {
+		logger.WithError(err).Error("Failed initialising tracing exporter")
```
rafaeljesus, line 153: we can use early return here and avoid the else below

pkg/jwt/handler.go
```diff
@@ -30,9 +30,10 @@ type Handler struct {
 // Reply will be of the form {"token": "<TOKEN>"}.
 func (j *Handler) Login(config config.Credentials) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		span := opentracing.FromContext(r.Context(), "token.extract")
+		_, span := trace.StartSpan(r.Context(), "token.extract")
```
rafaeljesus: We can get rid of the trace, it's a usual CPU bound operation, we are more concerned about network calls i.e IO bound

rafaeljesus: what we could actually do he needed is add annotate the existing tracing from the context with such information, however given it's request throughput is very low, I don't think it worth in the end

pkg/jwt/handler.go
```diff
@@ -41,9 +42,10 @@ func (j *Handler) Login(config config.Credentials) http.HandlerFunc {
 		factory := provider.Factory{}
 		p := factory.Build(r.URL.Query().Get("provider"), config)
 
-		span = opentracing.FromContext(r.Context(), "token.verify")
+		_, span = trace.StartSpan(r.Context(), "token.verify")
```
rafaeljesus: same as command above

pkg/jwt/handler.go
```diff
@@ -60,9 +62,10 @@ func (j *Handler) Login(config config.Credentials) http.HandlerFunc {
 			return
 		}
 
-		span = opentracing.FromContext(r.Context(), "token.issue_admin")
+		_, span = trace.StartSpan(r.Context(), "token.issue_admin")
 		token, err := IssueAdminToken(j.Guard.SigningMethod, claims, j.Guard.Timeout)
```
rafaeljesus, line 60: same here

pkg/jwt/handler.go
s4nji: Just for the record this trace was added in #270.  
I agree with the trace not being too useful, the metrics achieved the intended purpose of the PR; let's remove the trace(s).

vgarvardt: it was part of the auth problems investigation that is finished and it was not Janus issue at all but the calling-side problems

pkg/jwt/handler.go
rafaeljesus: just checked the code, here if can pass oc transport to github client, this way we can have traces and metrics for http calls

rafaeljesus: however I think it's too much for this PR, we can manage to improve later if you think it's take too much

rafaeljesus: apparently not too hard, we just need to inject the transport to GH client as follows:

```go
c := &http.Client{
  Transport: &ochttp.Transport{}
}

client := github.NewClient(c)
```

pkg/observability/observability.go
rafaeljesus: As a example @s4nji it will be as follows on dashboards:

*Incoming 99th Request Latency by Path Breakdown*
- `/api/menus/*` 500ms
- `/api/recipes/*` 100ms

When http client metrics enabled we would have the following:

*Outgoing 99th Request Latency by Path Breakdown*
- `http://api-v2.live-k8s.hellofresh.io/menus` 500ms
- `http://api-v2.live-k8s.hellofresh.io/recipes` 100ms

rafaeljesus: @spetrashov can you refer to @s4nji the OC guide-lines you wrote so he can expose the same metrics as other applications. Having that we can leverage the http://grafana.live-k8s.hellofresh.io/d/hellofresh-services/services-traffic dashboards to see services traffic

pkg/plugin/basic/handler.go
```diff
@@ -25,9 +25,9 @@ func NewHandler(repo Repository) *Handler {
 // Index is the find all handler
 func (c *Handler) Index() http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		span := opentracing.FromContext(r.Context(), "datastore.user.FindAll")
+		_, span := trace.StartSpan(r.Context(), "datastore.basic_auth.FindAll")
```
rafaeljesus: the trace span should be as close as possible to the real operation name `repo.FindAll`

pkg/plugin/oauth2/handlers.go
```diff
@@ -23,9 +23,9 @@ func NewController(repo Repository) *Controller {
 // Get is the find all handler
 func (c *Controller) Get() http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		span := opentracing.FromContext(r.Context(), "datastore.FindAll")
+		_, span := trace.StartSpan(r.Context(), "datastore.oauth_servers.FindAll")
```
rafaeljesus: same applies here inn regards to name span

pkg/jwt/handler.go
s4nji: `ochttp` is now used to collect http client stats and traces here:
https://github.com/hellofresh/janus/blob/ed29c5817f408b86a17e95b526d4baae57ff3c59/pkg/proxy/register.go#L62-L69

pkg/observability/observability.go
s4nji: Four new metrics are added via a9482e1, be38cb0, and 87ab607:
- `http_server_request_count_by_path`
- `http_server_latency_by_path`
- `http_client_request_count_by_path`
- `http_client_latency_by_path`

`path` for the server metrics refers to the `listen path` of the proxy definition. For the client metrics, `path` is the elected upstream when the request is proxied.

pkg/observability/observability.go
```diff
@@ -0,0 +1,98 @@
...
+	Stackdriver  = "stackdriver"
+	Zipkin       = "zipkin"
+)
+
+// PrometheusExporter is the prometheus exporter containing HTTP handler for "/metrics"
+var PrometheusExporter *prometheus.Exporter
+
+// Tags
+var (
+	KeyListenPath, _             = tag.NewKey("path")
```
s4nji, line 32: Should `path` be changed to `listen_path` to remove ambiguity?

cmd/init.go
```diff
@@ -80,3 +87,120 @@ func initStatsClient() {
...
+			ochttp.ClientResponseBytesView,
+			ochttp.ClientLatencyView,
+			ochttp.ClientRequestCountByMethod,
+			ochttp.ClientResponseCountByStatusCode,
+			ochttp.ServerRequestCountView,
+			ochttp.ServerRequestBytesView,
+			ochttp.ServerResponseBytesView,
+			ochttp.ServerLatencyView,
+			ochttp.ServerRequestCountByMethod,
+			ochttp.ServerResponseCountByStatusCode,
```
s4nji: Do we need all of these default views?

vgarvardt (approved):

spetrashov (approved):

PR #373, opened 2018-12-10
Fix config load returning nil pointers

s4nji:
Instead of returning nil pointers upon configuration load failure, return the configuration.

### DIFF
--- a/pkg/config/specification.go
+++ b/pkg/config/specification.go
@@ -7,6 +7,7 @@ import (
 	"github.com/kelseyhightower/envconfig"
 	"github.com/mitchellh/go-homedir"
 	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/config/specification.go
```diff
@@ -183,16 +184,14 @@ func Load(configFile string) (*Specification, error) {
...
 	// ensure the defaults are loaded
-	if err := viper.Unmarshal(&config); err != nil {
-		return nil, err
-	}
-
-	err := envconfig.Process("", &config)
+	err = viper.Unmarshal(&config)
 	if err != nil {
-		return nil, err
+		log.WithError(err).Warn("Failed unmarshaling config")
```
vgarvardt, line 192: what is the point of processing config of the unmarshalling failed?

s4nji: When unmarshalling fails due to invalid config on several lines, it may still return usable configuration; but I'm still not entirely sure if it is a good idea to keep running with invalid configuration.

Issue #359, opened 2018-09-18
StatsD Configuration


cchatfield:
Unable to get statsd to connect to a statsd hosts. The container is run and is pointed to a service running on the docker host.

Using the stats configuration I have tried the following to no avail:

ENV - STATS_DSN = "statsd://${NOMAD_IP_web}:8125"
STATS_PREFIX = "janus"

```2018/09/18 21:08:37 Could not initiale buffered client: address statsd://192.168.121.32:8125: too many colons in address. Falling back to a Noop Statsd client

{"error":"could not initialize statsd client: address statsd://192.168.121.32:8125: too many colons in address","event_name":"startup","level":"warning","msg":"an error occurred when an event was triggered","time":"2018-09-18T21:08:37Z"}
```

This ignores the IP.
ENV - STATS_DSN = "statsd:${NOMAD_IP_web}:8125"
STATS_PREFIX = "janus"

```
[StatsGo] 2018/09/18 21:12:42 Trying to connect to statsd instance	addr=	prefix=
[StatsGo] 2018/09/18 21:12:42 An error occurred while connecting to StatsD	addr=	prefix=	error=write udp 127.0.0.1:44623->127.0.0.1:8125: write: connection refused
{"error":"write udp 127.0.0.1:44623-\u003e127.0.0.1:8125: write: connection refused","level":"fatal","msg":"Error initializing stats client","time":"2018-09-18T21:12:42Z"}
```

What is the correct format to specify host and port?

**Janus version:** latest
**OS and version:** [Enter OS name and version here]

DineshReddyK:
Hi,
Yes, I am facing the similar issue. Init() seems to be fine as it says "Statsd metrics enabled", but later it fails with "too many colons" error.

```
janus[8779]: [StatsGo] 2019/02/08 20:30:39 Trying to connect to statsd instance        addr=127.0.0.1:8125        prefix=
janus[8779]: {"level":"debug","metrics_dsn":"statsd://127.0.0.1:8125","msg":"Statsd metrics enabled","plugin":"cb","plugin_event":"startup","time":"2019-02-08T20:30:39+02:00"}
janus[8779]: 2019/02/08 20:30:39 Could not initiale buffered client: address statsd://127.0.0.1:8125: too many colons in address. Falling back to a Noop Statsd client
[...]

PR #388, opened 2019-02-11
Fix circuit breaker plugin statsd collector

s4nji:
Fixes circuit breaker plugin using full dsn as address for statsd collector.

#### Reference
#359

### DIFF
--- a/pkg/plugin/cb/setup.go
+++ b/pkg/plugin/cb/setup.go
@@ -15,8 +15,7 @@ import (
 )
 
 const (
-	statsdPrefix = "hystrix"
-	pluginName   = "cb"
+	pluginName = "cb"
 )
 
 // Config represents the Body Limit configuration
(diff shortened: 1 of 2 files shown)


pkg/plugin/cb/setup.go
```diff
@@ -97,9 +100,15 @@ func onStartup(event interface{}) error {
...
 	logger.WithField("metrics_dsn", e.Config.Stats.DSN).Debug("Statsd metrics enabled")
+
+	dsnURL, err := url.Parse(e.Config.Stats.DSN)
+	if err != nil {
+		return errors.Wrap(err, "could not parse stats dsn")
+	}
+
 	c, err := plugins.InitializeStatsdCollector(&plugins.StatsdCollectorConfig{
-		StatsdAddr: e.Config.Stats.DSN,
-		Prefix:     statsdPrefix,
```
vgarvardt, line 102: is this const used anywhere else? if not - probably time to clean it out

vgarvardt (approved):

PR #403, opened 2019-07-19
Periodically clear http keep-alive connections

s4nji:
Adds a new optional environment variable `CONN_PURGE_INTERVAL` that periodically clears http keep-alive connections on the proxy.

This is done to limit the errors that occur when a dns change happens on a target host of an endpoint. If this endpoint is under constant load, the connection (with stale dns) will likely be used forever, as `IDLE_CONN_TIMEOUT` of 90 seconds will never be reached.

### DIFF
--- a/pkg/config/specification.go
+++ b/pkg/config/specification.go
@@ -17,6 +17,7 @@ type Specification struct {
 	MaxIdleConnsPerHost  int           `envconfig:"MAX_IDLE_CONNS_PER_HOST"`
 	BackendFlushInterval time.Duration `envconfig:"BACKEND_FLUSH_INTERVAL"`
 	IdleConnTimeout      time.Duration `envconfig:"IDLE_CONN_TIMEOUT"`
+	ConnPurgeInterval    time.Duration `envconfig:"CONN_PURGE_INTERVAL"`
 	RequestID            bool          `envconfig:"REQUEST_ID_ENABLED"`
 	Log                  logging.LogConfig
 	Web                  Web
--- a/pkg/proxy/register.go
+++ b/pkg/proxy/register.go
@@ -24,6 +24,7 @@ type Register struct {
 	router                 router.Router
 	idleConnectionsPerHost int
 	idleConnTimeout        time.Duration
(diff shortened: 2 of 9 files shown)


pkg/proxy/register.go
```diff
@@ -59,14 +60,13 @@ func (p *Register) Add(definition *RouterDefinition) error {
...
 	handler.FlushInterval = p.flushInterval
-	handler.Transport = &ochttp.Transport{
-		Base: transport.New(
-			transport.WithIdleConnTimeout(p.idleConnTimeout),
-			transport.WithInsecureSkipVerify(definition.InsecureSkipVerify),
-			transport.WithDialTimeout(time.Duration(definition.ForwardingTimeouts.DialTimeout)),
-			transport.WithResponseHeaderTimeout(time.Duration(definition.ForwardingTimeouts.ResponseHeaderTimeout)),
-		),
-	}
+	handler.Transport = transport.New(
```
vgarvardt: with this change we're losing transport metrics as I see

s4nji: The branch is still work in progress; we should still wrap transport with `ochttp`.

s4nji: Metrics are now back 🙌

pkg/config/specification.go
```diff
@@ -17,6 +17,7 @@ type Specification struct {
 	MaxIdleConnsPerHost  int           `envconfig:"MAX_IDLE_CONNS_PER_HOST"`
 	BackendFlushInterval time.Duration `envconfig:"BACKEND_FLUSH_INTERVAL"`
 	IdleConnTimeout      time.Duration `envconfig:"IDLE_CONN_TIMEOUT"`
+	ConnPurgeInterval    time.Duration `envconfig:"CONN_PURGE_INTERVAL"`
```
vgarvardt, line 20: this new setting must be documented

vgarvardt (approved):

vgarvardt:
👍 

Approved with Zappr  Approved with Zappr  Approved with Zappr

aleksandrzhiliaev:
+1

PR #407, opened 2019-08-06
Add debug trace

nenad:
## What does this PR do?
Adds DebugTrace feature which returns `X-Debug-Trace` response header with the Trace ID. 

Edit: I've reworked the PR so that now we match against some arbitrary key, instead of relying on `true` or `1` set on the request header.

### DIFF
--- a/janus.sample.toml
+++ b/janus.sample.toml
@@ -195,6 +195,11 @@
   #
   SamplingParam: "0.15"
 
+  # Requests that have X-Debug-Trace header set with the value of DebugTraceKey will
+  # force sampling, and additionally will have the Trace ID returned in their responses
+  # on the X-Debug-Trace header.
+  DebugTraceKey = ""
+
   [tracing.jaeger]
     # SamplingServerURL is the address to the sampling server
     #
--- a/janus.sample.toml
+++ b/janus.sample.toml
@@ -169,13 +169,13 @@
   #
   # Default: None
   #
-  Exporter: "jaeger"
+  Exporter = "jaeger"
 
   # Service name used in the backend
   #
   # Default: "janus"
   #
-  ServiceName: "janus"
+  ServiceName = "janus"
 
   # SamplingStrategy specifies the sampling strategy
   #
@@ -183,7 +183,7 @@
   #
   # Default: "probabilistic"
   #
-  SamplingStrategy: "probabilistic"
+  SamplingStrategy = "probabilistic"
 
   # SamplingParam is an additional value passed to the sampler.
   #
@@ -193,7 +193,7 @@
   #
   # Default: "0.15"
   #
-  SamplingParam: "0.15"
+  SamplingParam = "0.15"
 
   # Requests that have X-Debug-Trace header set with the value of DebugTraceKey will
   # force sampling, and additionally will have the Trace ID returned in their responses
@@ -205,4 +205,4 @@
     #
     # Default: None
     #
-    SamplingServerURL: "localhost:6832"
+    SamplingServerURL = "localhost:6832"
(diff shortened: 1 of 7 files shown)


pkg/middleware/debug_trace.go
```diff
@@ -0,0 +1,33 @@
...
+func DebugTrace(format propagation.HTTPFormat) func(handler http.Handler) http.Handler {
+	return func(handler http.Handler) http.Handler {
+		if format == nil {
+			format = &b3.HTTPFormat{}
+		}
+
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			debugVal := r.Header.Get(DebugTraceHeader)
+			if debugVal == "1" || debugVal == "true" {
+				ctx, span := trace.StartSpan(r.Context(), DebugTraceHeader, trace.WithSampler(trace.AlwaysSample()))
```
rafaeljesus, line 28: 👍

pkg/middleware/debug_trace.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"net/http"
+
+	"go.opencensus.io/plugin/ochttp/propagation/b3"
+	"go.opencensus.io/trace"
+	"go.opencensus.io/trace/propagation"
+)
+
+const (
+	DebugTraceHeader = "X-Debug"
```
vgarvardt: it should be more explicit, sothing like `X-Expose-Tracing-ID`

nenad: Does `X-Debug-Trace` make sense?

nenad:
@vgarvardt and @rafaeljesus I've updated the PR to be more strict on what is considered a valid debug request. I'd appreciate if you could review it once again.

pkg/config/specification.go
```diff
@@ -149,11 +150,12 @@ func init() {
 	viper.SetDefault("tracing.serviceName", serviceName)
 	viper.SetDefault("tracing.samplingStrategy", "probabilistic")
 	viper.SetDefault("tracing.samplingParam", 0.15)
+	viper.SetDefault("tracing.debugTraceKey", false)
```
vgarvardt: it is string now

pkg/middleware/debug_trace_test.go
```diff
@@ -0,0 +1,55 @@
...
+	}{
+		{
+			testName:            "'X-Debug-Trace: secret-key' produces response debug header",
+			debugHeader:         "secret-key",
+			expectedTraceHeader: true,
+		},
+		{
+			testName:            "'X-Debug-Trace: 0' does not produce response debug header",
+			debugHeader:         "0",
+			expectedTraceHeader: false,
```
vgarvardt: `"0"` is not an empty string

janus.sample.toml
```diff
@@ -193,11 +193,16 @@
   #
   # Default: "0.15"
   #
-  SamplingParam: "0.15"
+  SamplingParam = "0.15"
+
+  # Requests that have X-Debug-Trace header set with the value of DebugTraceKey will
+  # force sampling, and additionally will have the Trace ID returned in their responses
+  # on the X-Debug-Trace header.
+  DebugTraceKey = "secret-key"
```
vgarvardt: set it to empty value by default please

vgarvardt (commented):
couple minor issues and please add new feature to unreleased changes list https://github.com/hellofresh/janus/blob/master/CHANGELOG.md

rafaeljesus:
:shipit:

rafaeljesus:
int-test are failing

vgarvardt (approved):

PR #412, opened 2019-08-27
PT-5117 Fixed test to be compatible with go modules

vgarvardt:
## What does this PR do?
Since dependencies are handled with go modules - it is no longer required for package to be in the package name location.


rafaeljesus (approved):

rafaeljesus:
:shipit: 

PR #419, opened 2019-10-04
Respect parent sampling decision if not public endpoint

s4nji:
## What does this PR do?
Currently, if `IsPublicEndpoint` is set to false (and Janus does not ignore propagated traces), and the parent span decides that the trace should not be sampled, Janus still decides whether or not it should sample the trace.  
This causes traces without root span to be recorded if it decides to sample against the parent's decision.

This PR adds a new sampler that ensures sampling decision of the parent span are always respected. If there is no parent span (i.e. it's a direct request to Janus), only then the specified sampler will be used (which currently defaults to [`ProbabilitySampler(0.15)`][1]).

## Links and References
- https://github.com/hellofresh/opencensus-go-extras/blob/master/trace/sampling.go

[1]: https://github.com/hellofresh/janus/blob/63a0fc28d1c2e22c8a4aad03653a8112e95df4d1/pkg/config/specification.go#L152-L153

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 - None
 
+# 3.8.13
+- Changed trace sampling behavior when `TRACING_IS_PUBLIC_ENDPOINT` is set to false
+
 # 3.8.12
 
 ## Added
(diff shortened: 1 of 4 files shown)


nenad:
:+1:

Approved with Zappr  Approved with Zappr  Approved with Zappr

vgarvardt (approved):

s-radyuk:
+1

Approved with Zappr  Approved with Zappr  Approved with Zappr

PR #423, opened 2019-11-20
fix error on ingress or service name not specified

ashleyschuett:
## What does this PR do?
Fix template from erroring when name is omited


rafaeljesus (approved):

rafaeljesus:
:shipit: 

PR #442, opened 2020-08-11
EES-574 Support dynamic port for deps in integration tests

vgarvardt:
Right now it is not being used as travis allows static ports binding, but in GH Actions port is going to be dynamic, so it will be passed as env var to tests.

This PR adds dynamic port support to integration tests only, feature tests to be done.

One test marked as integration is simple unit test actually that does not require any deps, so build tag was removed.


aleksandrzhiliaev:
+1

RTodorov (approved):

PR #443, opened 2020-08-11
EES-574 Support dynamic port for deps in features tests

vgarvardt:
When running dependencies on non-local docker instance there is no way to mount volumes with mocks. So before running tests bootstrap creates all the mocks with API.

Since port is going to be dynamic - we need to support this thing as well.


fftorres:
+1

RTodorov (approved):

PR #448, opened 2020-08-17
EES-574 Use ubuntu for docker image and run as nobody

vgarvardt:


s4nji (approved):
+1 lgtm

fftorres:
+1

Dockerfile
```diff
@@ -1,30 +1,33 @@
 FROM golang:1.14-alpine AS builder
```
boekkooi-fresh, line 1: Please don't mix ubuntu for run and alpine for build it's known to cause errors.

vgarvardt: It will go away in the next PR. This one was only about changing runtime. Btw, `CGO_ENABLED=0` kinda solves a problem with building/running in glibc/musl-libc envs.

PR #450, opened 2020-08-17
EES-574 Ignore freebsd/arm64 build as it can not be done

vgarvardt:
Adding to ignore because of

```
   ⨯ release failed after 369.34s error=failed to build for freebsd_arm64: # golang.org/x/sys/unix
../../../go/pkg/mod/golang.org/x/sys@v0.0.0-20190825160603-fb81701db80f/unix/ztypes_freebsd_arm64.go:400:12: undefined: uint128
```

All other builds seems yo be working, at least locally.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -22,6 +22,8 @@ builds:
     ignore:
       - goos: darwin
         goarch: 386
+      - goos: freebsd
+        goarch: arm64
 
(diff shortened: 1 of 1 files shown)


s4nji:
+1

Approved with Zappr  Approved with Zappr

s4nji (approved):

PR #457, opened 2020-09-02
EES-3845 Updated health-go to v3

vgarvardt:
Use latest stable version of health-go lib


boekkooi-fresh:
:+1: 

main.go
```diff
@@ -13,7 +11,6 @@ func main() {
 	rootCmd := cmd.NewRootCmd(version)
 
 	if err := rootCmd.Execute(); err != nil {
-		log.WithError(err).Error(err.Error())
-		os.Exit(1)
```
boekkooi-fresh, line 17: Personally I prefer the explicit exit since a logger is often an interface and has no guarantee that it will exit.
In this case I see no issue.

fftorres:
+1

RTodorov (approved):

PR #458, opened 2020-09-02
EES-3759 Updated redis and limiter libraries 

vgarvardt:
Current deps are very outdated, update some one-by-one. In this case redis and limiter updated at once because one depends on another.


aleksandrzhiliaev:
+1

RTodorov (approved):

PR #459, opened 2020-09-03
EES-3845 Use official mongo driver

vgarvardt:
Official mongo driver is available for pretty long time. Would be nice to start using it instead of community one.


pkg/plugin/events.go
```diff
@@ -1,12 +1,13 @@
 package plugin
 
 import (
-	"github.com/globalsign/mgo"
+	"github.com/hellofresh/stats-go/client"
+	"go.mongodb.org/mongo-driver/mongo"
```
iammuho, line 5: Wrong order

iammuho:
+1

Approved with Zappr  Approved with Zappr

aleksandrzhiliaev:
+1

RTodorov (approved):

PR #485, opened 2022-11-21
Improve access rule middleware

nox-4O4:
This PR improves the access rule middleware:
- bugfix: a non-matching deny rule wrongly blocks the request when evaluated a second time
- bugfix: handler function gets executed multiple times when allow rules match
- improvement: do not ignore multiple rules
- housekeeping: simplify tests for access rule middleware

### DIFF
--- a/pkg/plugin/oauth2/middleware_access_rules_test.go
+++ b/pkg/plugin/oauth2/middleware_access_rules_test.go
@@ -15,110 +15,6 @@ import (
 
 const signingAlg = "HS256"
 
-func TestBlockJWTByCountry(t *testing.T) {
-	secret := "secret"
-
-	revokeRules := []*AccessRule{
-		{Predicate: "country == 'de'", Action: "deny"},
-	}
-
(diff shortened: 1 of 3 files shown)


pkg/plugin/oauth2/middleware_access_rules_test.go
```diff
@@ -15,174 +15,129 @@ import (
 
 const signingAlg = "HS256"
 
-func TestBlockJWTByCountry(t *testing.T) {
-	secret := "secret"
+func generateToken(alg, key string) (string, error) {
+	token := basejwt.NewWithClaims(basejwt.GetSigningMethod(alg), basejwt.MapClaims{
+		"country":  "de",
+		"username": "test@hellofresh.com",
```
ericklb84, line 21: Mocked info?

nox-4O4: Yes, those are mocked / static claims for testing the access rules. (They were previously in between test cases, I moved them to the top.)

